PR #17, opened 2015-02-15
Adding support for ValueLink, minHeight / maxHeight, & reusability

eugene1g:
Hi,

I started this to add support for ValueLink, but then imported other changes that might be useful to the rest of the community. It's fully backwards compatible, but has couple additions -

1. Works with React's ValueLink
2. Introduced minHeight & maxHeight properties (set as numbers)
3. Manages 'height' and 'overflow' styling within React's state to make sure they are persisted if the component is re-rendered from the outside

Tested on IE9+. Any feedback is most welcome!

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,79 +4,153 @@ var React = require('react');
 var objectAssign = require('object-assign');
 
 var TextareaAutosize = React.createClass({
-  displayName: 'TextareaAutosize',
+    displayName: 'TextareaAutosize',
 
-  render: function() {
-    var props = objectAssign({}, this.props, {
-      onChange: this.onChange,
-      style: objectAssign({}, this.props.style, {overflow: 'hidden'})
-    });
+    propTypes: {
+        minHeight: React.PropTypes.number,
+        maxHeight: React.PropTypes.number
+    },
 
-    return React.DOM.textarea(props, this.props.children);
-  },
+    /**
+     * This component can work with the ValueLink shortcut provided by React.
+     * When ValueLink is passed, it tries to take over the onChange event in the underlying element
+     * However, we need onChange to calculate the height of the box ourselves
+     * So we store ValueLink locally, remove it from the underlying <textarea>, but proxy all onChange calls to it
+     */
+    valueLink: undefined,
 
-  componentDidMount: function() {
-    this.recalculateSize();
-    window.addEventListener('resize', this.recalculateSize);
-  },
+    /**
+     * A cached copy of properties that will be passed to the underlying <textarea>
+     * Essentially passes all own properties, but sanitizes them to remove ValueLink and add the necessary style settings
+     */
+    textareaProps: {},
 
-  componentWillUnmount: function() {
-    window.removeEventListener('resize', this.recalculateSize);
-  },
+    getInitialState: function () {
+        return {
+            height:   'auto',
+            overflow: 'hidden'
+        }
+    },
 
-  componentDidUpdate: function(prevProps) {
-    if (
-      prevProps.style ||
-      prevProps.value !== this.props.value ||
-      // We need to call `recalculateSize()` if component is used in
-      // uncontrolled mode because the `<textarea />` will re-render itself in
-      // that case.
-      //
-      // We check for `== null` to cover both `null` and `undefined`, this is
-      // what React does to check if component is used in uncontrolled mode.
-      this.props.value == null
-    ) {
-      this.recalculateSize();
+    render: function () {
+        this.textareaProps.style = objectAssign({}, this.textareaProps.style, {
+            height:   this.state.height,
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Hey @eugene1g read through this just now and there's a ton of great stuff here! Thanks!

Would you mind though splitting this PR into separate PRs? The ValueLink stuff I definitely want to merge in but the other changes I don't understand as well and it's easier to evaluate multiple changes in isolation than when they're all together.

index.js
```diff
@@ -4,79 +4,153 @@ var React = require('react');
...
+
+
+    /**
+     * Do the initial calcluation of the height
+     * And set an event listener to update the textarea height when the window is resized
+     */
+    componentDidMount: function () {
+        //do a short timeout so let the browser sort out its paddings etc
+        //otherwise sometimes the browser doesn't not text should be wrapped and missed some cases
+        setTimeout(this.recalculateSize);
```
nikgraf: hey @andreypopp & @eugene1g - I'm building a similar Input & found your code very helpful. Thanks for the great work!

@eugene1g If you use setTimeout doesn't it sometimes make the whole interface jumping around, because for a brief period the height is not set properly?

eugene1g: 1ms should be quick enough for the jump not to be noticeable, but sometimes browsers misbehave with timeout and yes you might see a jump.

This particular implementation is quite naive because it directly manipulates DOM, and that poses two problems: page scrolling is affected and it is not very React-idiomatic. Additionally there is a performance hit from calculating computedStyle on each key stroke.

I didn't end up using this code and wrote another one specifically fitted to our needs, but didn't have a chance to prepare multiple PR for this project yet. I'll post up our version shortly for you to review & compare.

nikgraf: Thx, looking forward to it!

eugene1g: For now I put up the relevant pieces at https://gist.github.com/eugene1g/5dbaa7d35d0c7d5c2c56 as an implementation reference.

Couple disclaimers:
1. The code is structured for ES6 and React 0.13, so you may need to have it transpiled (we use babel across all scripts so that's not an issue here)
2. There are two files - one is React-specific that renders the component, and the other one is DOM-specific that measures how long the text should be. This is because we still have old-school DOM interactions elsewhere in the app, and need to have the measuring logic available elsewhere. But it could be rolled up into the same component
3. The measuring script injects an additional invisible textarea into the DOM for all of its work and styles it accordingly to make sure its height measurements are the same as the real textarea node. This is necessary because adjusting the height on the React dom node directly often ends up messing up the scrolling position on the page.
4. For performance reasons, the DOM measuring util caches all relevant CSS descriptors of the textarea. This works well for us, but if you're doing dynamic manipulation of paddings / line-heights etc then you may need to bust that cache or remove it altogether.

nikgraf: This is fantastic! Thank you very much. Especially for putting in the effort & explaining your decisions here.

andreypopp: @eugene1g thanks for the example code! I think it's worth incorporating this into react-textarea-autosize.

KyleAMathews: Second what @andreypopp said. Looks great. And I've noticed the problems with page scrolling with the current code that you allude to. If you want to put together a PR request that'd be fantastic. I'm perfectly fine with ES6 code if that's what you'd like.

eugene1g: Hey guys,

I will try to revise the PR, but for now I've put up our current implementation to https://gist.github.com/eugene1g/6547c28997f318924c89 (in case someone has a spare hour to rebase)

nikgraf: @eugene1g @andreypopp & @KyleAMathews Thanks again for sharing your great work. I now implemented a TextInput component that allows autosizing & switch between allowNewLine true/false.

Here is the code: https://github.com/nikgraf/belle/blob/master/src/components/TextInput.jsx
Here you can try it out: http://nikgraf.github.io/belle/#text-input

The library is still early stage, but I would be happy about feedback :)

nikgraf: @eugene1g I forked your Gist & added a fix for Firefox & some improvements for height calculations.

eugene1g: Great to hear that! I can't find the fork anywhere within my github references - was it with this commit https://github.com/nikgraf/belle/commit/c6ce4b6e33550e34b814e97436f8e8c0199be1d3 ?

nikgraf: Here you can find the latest version: https://gist.github.com/nikgraf/c4acdfa9659d8219b916/75af1530669746a2e6342f46c1aa04b9cd5cb8c4

By clicking on the revisions on the right of a gist you can see the different versions: https://gist.github.com/nikgraf/c4acdfa9659d8219b916/revisions

andreypopp:
Fixed by #19

andreypopp:
Thanks @eugene1g !

PR #19, opened 2015-03-27
New implementation

### LINKED ISSUE #12: Doesn't work with React.addons.LinkedStateMixin
I get `Uncaught Error: Invariant Violation: Cannot provide a valueLink and a value or onChange event. If you want to use value or onChange, you probably don't want to use valueLink.` Makes sense, but it would be nice if this worked with the LinkedStateMixin as a "drop-in replacement".

andreypopp:
This is the implementation made by @eugene1g (see #17), fixes #12.

- Use ghost <textarea /> attached to body to calculate height
- Optionally cache DOM measurements
- ES6/7 syntax
- React >= 0.13

Todo:

- [ ] Add tests
- [ ] Recalc size on window resize
- [ ] Add lint (via eslint)


andreypopp:
cc @eugene1g @KyleAMathews

eugene1g:
@nikgraf has made additional improvements (https://gist.github.com/nikgraf/c4acdfa9659d8219b916/revisions)  to fix a memory leak issue and I think his changes should be included in the new version

nelix:
This is very nice indeed.

andreypopp:
Updated PR with changes pointed by @eugene1g.

src/TextareaAutosize.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  static defaultProps = {
+    onChange: emptyFunction,
+    useCacheForDOMMeasurements: false
+  }
+
+  constructor(props) {
+    super(props);
+    this.props = props;
```
nikgraf: Isn't this.props = props be called by super(props)?

src/TextareaAutosize.js
```diff
@@ -0,0 +1,113 @@
...
+    props.style = {
+      ...props.style,
+      overflow: 'hidden',
+      height: this.state.height
+    };
+    return <textarea {...props} onChange={this._onChange} />;
+  }
+
+  componentDidMount() {
+    // Timeout seems to be required for all media queries to be applied
```
nikgraf: Do you have simple example to verify the issue?

KyleAMathews:
Was playing with this tonight as part of a side project. Seems solid.

src/TextareaAutosize.js
andreypopp: Yes, fixed.

src/TextareaAutosize.js
andreypopp: Yes, and... it depends on order of stylesheets vs js code not really on if style are inside media query rules or not.

So if we have
```
    <link rel="stylesheet" href="bundle.css" />
    <script src="bundle.js"></script>
```
then we see correct values in componentDidMount, but in case of
```
    <link rel="stylesheet" href="bundle.css" />
    <script src="bundle.js"></script>
```
we need to wait some time before measuring DOM.

I think we can remove `onNext` call in `componentDidMount` and require users to re-render in case any of styles on page which can influence textarea sizing changed.

andreypopp:
Updated PR

src/TextareaAutosize.js
nikgraf: @andreypopp What is the difference between the two code snippets?

src/TextareaAutosize.js
nikgraf: :+1:

src/TextareaAutosize.js
andreypopp: We see correct sizings in componentDidMount or not (because styles are not loaded yet). That's unrelated to media queries though.

nikgraf:
@andreypopp @KyleAMathews @eugene1g I noticed to calculate the height properly for a textarea with `box-sizing: border-box` you also need include `border-width` & `box-sizing`

Maybe you can cherry pick it: https://github.com/nikgraf/react-textarea-autosize/commit/7742b12832ac1faaedf8fbe3cf619bd6fd61f3da

andreypopp:
@nikgraf added you as a collaborator, you can push your commits to this branch

nikgraf:
@andreypopp thx, just pushed :smile:

andreypopp:
@nikgraf thanks

OK, merging it in.

KyleAMathews:
thanks @eugene1g @nikgraf and @andreypopp! Very nice update.

nikgraf:
@KyleAMathews my pleasure :+1: 

should we delete the branch or do you want to keep it for some reason?

KyleAMathews:
@nikgraf @andreypopp deleted it

PR #22, opened 2015-04-14
Fix screen jumpness when textarea is taller than browser window

tpcstld:
When the textarea is taller than the browser, the screen jumps occasionally when you try to edit the textarea's content. This PR fixes that issue.


KyleAMathews:
@tpcstld have you tried @andreypopp PR #19? I believe it has a fix for the same issue. I'm going to review it soon and it'd be great to get your review of the new implementation as well so we can get it released.

andreypopp:
I believe it is fixed by #19

PR #26, opened 2015-05-14
allow styles to change overflow

tomatau:
E.g.

I have a textarea for editing that needs autosize but also is affixed.

When the content is too long I set a maxHeight so that the affix doesn't drop off the bottom of the page.. but now I wish to set `overflow: auto` so that textarea can be scrolled.

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -47,8 +47,8 @@ export default class TextareaAutosize extends React.Component {
       props.value = this.props.valueLink.value;
     }
     props.style = {
-      ...props.style,
       overflow: 'hidden',
+      ...props.style,
       height: this.state.height
     };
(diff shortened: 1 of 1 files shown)


andreypopp:
I'd prefer not to set `overflow: hidden` when we have `maxHeight` or `rows` set and content overflows the textarea. That way you request would be fulfilled and it won't require manual style fixing.

tomatau:
I don't follow? Right now the autosized-textarea **is** setting `overflow: hidden` which cannot be changed by the implementation.

Assuming the `overflow: hidden` is there for a reason? Might it not be more simple to just allow implementations to explicitly override this hidden setting instead of having the textarea respond to edge cases?

andreypopp:
Fix released in `2.1.3`.

PR #27, opened 2015-05-15
Check if we're on server or not

alexfedoseev:
This check fixes server errors, where `window` is not defined.

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -98,10 +98,14 @@ export default class TextareaAutosize extends React.Component {
 
 }
 
-let onNextFrame = window.requestAnimationFrame;
+if (process.browser) {
 
-if (onNextFrame === undefined) {
-  onNextFrame = function onNextFrame(cb) {
-    window.setTimeout(cb, 1);
-  };
+  let onNextFrame = window.requestAnimationFrame;
+
+  if (onNextFrame === undefined) {
+    onNextFrame = function onNextFrame(cb) {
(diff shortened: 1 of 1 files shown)


src/TextareaAutosize.js
```diff
@@ -98,10 +98,14 @@ export default class TextareaAutosize extends React.Component {
 
 }
 
-let onNextFrame = window.requestAnimationFrame;
+if (process.browser) {
```
andreypopp: Does it work with webpack? I think more reliable would be to move check for `requestAnimationFrame` within the `onNextFrame` function.

alexfedoseev:
Using it with webpack, works fine.

src/TextareaAutosize.js
alexfedoseev: Using it with webpack, works fine.

alexfedoseev: Or we can make check this way:

```javascript
if (typeof window !== 'undefined') {
  // ...
}

andreypopp: In case `componentDidMount` will execute in window-less environments (unlikely but probable) that will make the error message confusing.

I moved code which reference `window` to within the `onNextFrame` function so error message will mention `window`. This also prevents previous lint failures due to referencing undefined `onNextFrame`.

I released your changes in `2.1.2`. Thank you.

alexfedoseev: Updated, everything smooth in both envs. Thanks!

Issue #30, opened 2015-06-01
Uncaught Error: Invariant Violation


alexfedoseev:
Further to #24 

Here is the sample app to reproduce an issue: https://github.com/alexfedoseev/flummox-cuts
Clone repo && `npm install` && `npm start`.

If you'll click on any tag link in the post list, there will be an error in console:

![image](https://cloud.githubusercontent.com/assets/4244251/7907131/26107676-083e-11e5-807b-4c147598bb0f.png)

andreypopp:
Released 2.2.3 with fixes by @eugene1g, it should fix this issue.

alexfedoseev:
@andreypopp @eugene1g Everything's ok, thanks!

PR #31, opened 2015-06-01
(Please review) Cancelling scheduled events

eugene1g:
Under some circumstances, _resizeComponent can get called after the node is unmounted (because the action was scheduled via onNextFrame). This change attempts to remove all listeners when the element is unmounted which should prevent that error (see #30).

Sorry I don't have a way to test these changes as I'm travelling without my gear.


andreypopp:
Looks good, thanks!

PR #32, opened 2015-06-05
fix height calculation for box-sizing:border-box with a border

jenseng:
this fixes an issue where the height would always be a little too small if you have a border, resulting in perma-scrollbars in many browsers

incidentally this now also supports box-sizing:padding-box (firefox only)


eugene1g:
Good call - I used CSS elsewhere to suppress overflow but factoring those pixels in is much cleaner

andreypopp:
Thanks

PR #34, opened 2015-06-09
fix border-related calculations in some browsers

jenseng:
this broke for some browsers in 9b65e4d2, since they just return an empty string for border-top/border-bottom... sorry about that, should have tested the previous commit in more than just chrome :(


andreypopp:
Thanks, published as `2.3.1`.

jenseng:
awesome, thanks :+1:

PR #38, opened 2015-07-07
Invoke callback on height changes.

### LINKED ISSUE #36: Invoke callback when textarea changes it's height.
It would be great if there is an option to pass callback which is invoked when text area changes it's height.

Andreyco:
Implements and closes #36

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -18,6 +18,11 @@ export default class TextareaAutosize extends React.Component {
      * Callback on value change.
      */
     onChange: React.PropTypes.func,
+    
+    /**
+     * Callback on height changes.
+     */
+    onHeightChange: React.PropTypes.func,
 
     /**
      * Try to cache DOM measurements performed by component so that we don't
@@ -46,6 +51,7 @@ export default class TextareaAutosize extends React.Component {
 
   static defaultProps = {
     onChange: emptyFunction,
+    onHeightChange: emptyFunction,
     useCacheForDOMMeasurements: false
   }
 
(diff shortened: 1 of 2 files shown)


src/TextareaAutosize.js
```diff
@@ -87,6 +93,13 @@ export default class TextareaAutosize extends React.Component {
     // Re-render with the new content then recalculate the height as required.
     this.onNextFrameActionId = onNextFrame(this._resizeComponent);
   }
+  
+  componentWillUpdate(nextProps, nextState) {
```
chrislloyd: Is this the best place to invoke the callback? It'll block the update until the callback completes.

andreypopp: Also what if parent decides to re-render (calls its `setState` from within the callback)?

Andreyco: @chrislloyd Probably invoke callback from *componentDidUpdate*?
@andreypopp yes, it can lead into endless loop. Do you have any idea, how to prevent this?

andreypopp: @Andreyco I'm more about the fact that `setState` is not allowed from within the `componentWillUpdate` so it is interesting, what if parent is updated from there and re-rendered its children including this component.

andreypopp: Yes, I think `componentDidUpdate` is the right place here.

Andreyco: I agree with you calling `setState` from within componentWillUpdate will cause you troubles. 
Modifying Textarea's height from parent's callback will do the same. It's up to developer to use the callback properly, right?

andreypopp: @Andreyco regarding the infinite loop due to parent's change to textarea height, I agree.

Andreyco: from time to time, everyone burns the fingers ;)

andreypopp:
Thanks! Released in 2.4.0.

PR #40, opened 2015-07-24
clear next frame action before scheduling another

bluej100:
This resolves an error we were seeing in our unit tests.
Two rapid prop changes followed by an unmount would cause
an error, since only the second would be cancelled.


andreypopp:
Thanks.

bluej100:
Would you mind doing a patch release?

andreypopp:
@bluej100 released in 2.4.1

bluej100:
Thanks! Appreciate your time.

PR #42, opened 2015-08-06
Support all React 0.14 betas and alphas

frederickfogerty:
This module is working fine on React 0.14 for me, I think this should be added as a dep


andreypopp:
Thanks

PR #44, opened 2015-08-14
React 0.14.0 compat

### LINKED ISSUE #43: Broken in React 0.14.0 beta3
`react/lib/emptyFunction` does not exist in react 0.14.0 beta3

rendro:
Fixed #43 

* removed `react/lib/emptyFunction` dependency (seems to be deprecated in 0.14.0-beta3)
* added `react-dom` for `findDOMNode`
* bumped peer-dependency versions

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -105,8 +105,8 @@ export default class TextareaAutosize extends React.Component {
   }
 
   componentWillUnmount() {
-    //remove any scheduled events to prevent manipulating the node after it's
-    //been unmounted
+    // remove any scheduled events to prevent manipulating the node after it's
+    // been unmounted
     this.clearNextFrame();
   }
 
@@ -137,6 +137,7 @@ export default class TextareaAutosize extends React.Component {
 
   /**
    * Read the current value of <textarea /> from DOM.
+   * @return {String} value of <textarea />
    */
   get value(): string {
     return ReactDOM.findDOMNode(this).value;
@@ -144,6 +145,7 @@ export default class TextareaAutosize extends React.Component {
 
   /**
    * Put focus on a <textarea /> DOM element.
+   * @returns {undefined} No return value
    */
   focus() {
     ReactDOM.findDOMNode(this).focus();
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -3,8 +3,10 @@
  */
 
 import React from 'react';
-import emptyFunction from 'react/lib/emptyFunction';
+import ReactDOM from 'react-dom';
 import calculateNodeHeight from './calculateNodeHeight';
+
+const emptyFunction = () => {};
 
 export default class TextareaAutosize extends React.Component {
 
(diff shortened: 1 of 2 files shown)


src/TextareaAutosize.js
```diff
@@ -3,9 +3,11 @@
  */
 
 import React from 'react';
-import emptyFunction from 'react/lib/emptyFunction';
+import ReactDOM from 'react-dom';
```
andreypopp: This makes it incompatible with React 0.13, right?

blittle:
:+1:

jquense:
So we have been using our fork with the same changes as this PR, and are noticing that when typing in the middle of the textarea the cursor will suddenly jump to end of the textarea after the first key stroke. Needless to say its not a helpful behavior. Still trying to track down what the cause is, since it doesn't occur in react 0.13

src/TextareaAutosize.js
rendro: I guess so, but I am not entirely sure. I theory it should work with older versions but you end up with a duplicate findDOMNode implementation in both react 0.13 and react-dom.

jquense: `findDOMNode` depends on a bunch of stateful stuff so including it twice will blow up. This effectively makes it impossible for libraries to support both versions at once. That being said if someone using 0.13 was using webpack (or you provide a built bundle) then it could be made to work fine, with aliases

rendro: Thanks @jquense for the insights. So how should we handle this? I guess the only reasonable way to go is to wait until react and react-dom 0.14 is stable and then merge the PR and remove the support for 0.13. Any other ideas?

KyleAMathews: We could just make a new major release.

blittle: :+1: That is what you will do once React 0.14 releases anyway

koistya:
Mm... also you may want to add `"files": ["src/"]` to `package.json`

andreypopp:
@koistya why? `src` is explicitly in `.npmignore` and there's `lib` instead with compiled es5 code.

koistya:
This way it could be possible to install this library from github (e.g. in case you need to fork it, make some changes and install from GitHub instead of NPM)

andreypopp:
@koistya probably we should add some hooks for npm to compile dep after install.

src/TextareaAutosize.js
```diff
@@ -127,7 +129,7 @@ export default class TextareaAutosize extends React.Component {
   _resizeComponent() {
     let {useCacheForDOMMeasurements} = this.props;
     this.setState(calculateNodeHeight(
-      React.findDOMNode(this),
+      ReactDOM.findDOMNode(this),
```
andreypopp: I think can make it work completely without `react-dom` package if we set a `ref` on component's DOM node which in React 0.14 will return a DOM node. So we don't need `ReactDOM.findDOMNode` at all.

koistya: :+1:

andreypopp:
Ref c72b0b77229aae775de11c91257ca8ffb3558aa5

PR #46, opened 2015-08-19
fix(calculateNodeHeight): prevent return 0 for empty textareas

nikgraf:
If I remember correctly we did this change in Belle, because some IE versions the height will have scrollHeight of `0` in case the textarea has no value. I can double check this change tomorrow with a Windows computer.


andreypopp:
Thank you!

PR #51, opened 2015-09-14
Add text-rendering CSS property

razh:
Using `text-rendering: optimizeLegibility` enables kerning and optional
ligatures, which may affect text metrics.

Here's an example of a pathological case of kerning:

##### `text-rendering: auto` (default)
screen shot 2015-09-14 at 6 30 59 pm

##### `text-rendering: optimizeLegibility`
screen shot 2015-09-14 at 6 31 11 pm

### DIFF
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -20,6 +20,7 @@ const SIZING_STYLE = [
   'font-family',
   'font-weight',
   'font-size',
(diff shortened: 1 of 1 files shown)


nikgraf:
Looks good! Thanks for your contribution.

andreypopp:
Nice! Thank you!

Issue #47, opened 2015-09-01
Why is babel/polyfill required?


KyleAMathews:
I've forgotten why.

andreypopp:
Hm... I added these lines in the README but can't remember either...

andreypopp:
Removed the line.

Issue #53, opened 2015-09-19
react 0.14.0-rc1


growlsworth:
wanted to check to see if the package.json could be updated to support react 0.14

andreypopp:
`package.json` allows react 0.14.0-rc1

PR #56, opened 2015-09-21
Updating to work with React 0.14.0-rc1

nickpresta:
* Updated peerDependencies
* Updated render method to not mutate props
* Removed bit about babel polyfill

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -68,22 +68,22 @@ export default class TextareaAutosize extends React.Component {
   }
 
   render() {
-    let {valueLink, onChange, ...props} = this.props;
-    props = {...props};
+    let {valueLink, onChange} = this.props;
+    let mergedProps = {...this.props};
     if (typeof valueLink === 'object') {
-      props.value = this.props.valueLink.value;
+      mergedProps.value = mergedProps.props.valueLink.value;
     }
-    props.style = {
-      ...props.style,
+    mergedProps.style = {
+      ...mergedProps.style,
       height: this.state.height
     };
     let maxHeight = Math.max(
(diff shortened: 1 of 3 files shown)


nickpresta:
This should fix https://github.com/andreypopp/react-textarea-autosize/issues/47 and https://github.com/andreypopp/react-textarea-autosize/issues/53

tomatau:
I've forked this as well and am tracking built lib so that can install directly from github repo. Also working in 0.14-rc1
```json
"dependencies": {
"tomatau/react-textarea-autosize#react-0.14",
```

src/TextareaAutosize.js
```diff
@@ -68,22 +68,22 @@ export default class TextareaAutosize extends React.Component {
   }
 
   render() {
-    let {valueLink, onChange, ...props} = this.props;
-    props = {...props};
+    let {valueLink, onChange} = this.props;
+    let mergedProps = {...this.props};
```
andreypopp: Why this change?

andreypopp:
Closing in favor of #44 which was submitted first and appears to be more targeted towards a single change (updating codebase for react 0.14).

nickpresta:
@andreypopp if you don't update the props to make a copy and use that, you get errors about mutating props (which was discouraged in React 0.13):

```
ERROR: 'Warning: `textarea` was passed a style object that has previously been mutated. Mutating `style` is deprecated. Consider cloning it beforehand. Check the `render` of `TextareaAutosize`. Previous style: {"height":null}. Mutated style: {"height":null}.'
```

andreypopp:
@nickpresta that's interesting... at kind 72 code already clones `props` object.

PR #57, opened 2015-09-26
resizeComponent should listen to window resize events

kakadiya91:
Textarea component should automatically adjust its height when we resize browser window.
Changes:
- resizeComponent should listen to window resize events


andreypopp:
LGTM

PR #64, opened 2015-10-16
improve height calculations to not be influenced by global css

nikgraf:
@andreypopp any downsides I might have missed?


andreypopp:
Thanks!

PR #65, opened 2015-10-19
Ability to set value

jtadmor:
Hi,

Thanks for writing this library. I added in the small feature of being able to programmatically set the value of the underlying <textarea /> component using the same syntax and same implementation that you were using for getting the value.


andreypopp:
Thanks!

Issue #54, opened 2015-09-19
What is the use case for `onHeightChange` callback?


necolas:
(Context: I am playing around with using this in https://github.com/necolas/react-native-web/)

What do you see as the use case for this? The callback takes a `height`, which seems to be the CSS height (respects `border-box`) rather than the layout height of the element.

So to implement `onLayout` I need to re-query the DOM anyway, to get the outer height/width/offset. But one thing I cannot do easily is infer the number of rows. That might be quite useful to pass to the callback.

andreypopp:
Sounds reasonable. i'd accept a PR for that.

necolas:
I'll probably drop this dependency as it's not actively developed or tested, so won't get around to adding this change.

PR #72, opened 2015-11-17
Make rowCount available to onHeightChange callback

tkh44:
For #54

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -141,13 +141,13 @@ class Demo extends React.Component {
 {`
   <TextareaAutosize
     useCacheForDOMMeasurements
-    onHeightChange={height => console.log(height)}
+    onHeightChange={(height, rowCount) => console.log(height, rowCount)}
     />
 `}
           </pre>
           <TextareaAutosize
             useCacheForDOMMeasurements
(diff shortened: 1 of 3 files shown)


colllin:
Putting `rowCount` as a second parameter to the `onHeightChange` callback feels a little arbitrary, doesn't it?  What about passing an object as the second parameter, with a `rowCount` property?  Then we have a reasonable way of sending back other properties too, such as `outerHeight`.  Maybe it would be even better to create a separate callback instead of `onHeightChange`, for example `onAutosize`, which receives a single `data` parameter which contains `{height, rowCount, outerHeight}`.  Are there other uses besides `rowCount` and `outerHeight`?  Anything other height/size-related properties that someone might need to query the DOM for unless we return it here?

Andarist:
@tkh44 @colllin other option would be to expose `this` as second param and ofc keep `rowCount` on the instance. I think many other libs do this? WDYT? Second param as `extraData` seems reasonable too.

@tkh44 do you want to work a little bit so this can get merged in or should I pick it up?

tkh44:
@Andarist go ahead

Issue #74, opened 2015-12-18
Not working in ie11 - wrong height


vavdav:
Hello,
I have problem with textarea height in ie11.  
Problem is on this line:
https://github.com/andreypopp/react-textarea-autosize/blob/master/src/calculateNodeHeight.js#L73

IE11 is returning 2. Another browser is returning 21. 

Please how to solve this problem?

mitchcunha:
I saw something similar with IE11, but the right height was being calculated, but the state wasn't being set correctly.

The temporary fix that I found was changing the setState call in TextareaAutosize.js on line 128 to this:

this.setState(_calculateNodeHeight2'default');

It had some other || parameters, but that was causing some bugginess in IE11. Hope this helps, or if you know about this already, please let me know!

Best,

Kevin

CoderK:
You can find the solution in the link below.
[https://github.com/andreypopp/react-textarea-autosize/tree/fix-ie-height-calculation]

vavdav:
It is not working. It is happening when I have already some text inside textarea. I am using version 3.2.0

vavdav:
It is not working with '\uFEFF' but with 'x' it is working

just-paja:
Why don't we use '\u00A0' aka NBSP aka ' '?

dshorowitz:
I just came across this problem as well. In fact, you can see the issue by going to the demo page (http://andreypopp.github.io/react-textarea-autosize/). In IE11, as you start typing in the blank textarea, its height grows too early. Anyone have a fix? Thanks :)

jdalton:
Fixed https://github.com/andreypopp/react-textarea-autosize/issues/55#issuecomment-242201491.

PR #77, opened 2016-01-04
fix wrong scrollHeight value in IE11

vavdav:


vavdav:
related to https://github.com/andreypopp/react-textarea-autosize/issues/74

andreypopp:
Did you test against FF, we changed `'x'` to `''` to solve a bug in FF. See https://github.com/andreypopp/react-textarea-autosize/commit/88de2d32875ded4fba4a7c40be12dd7c5f1207fb

vavdav:
It behaves same with '' and with ' ' in example. Not sure how to reproduce it? And how to fix ie11 then?

dshorowitz:
@vavdav When you closed this, did you find a fix?

PR #80, opened 2016-01-14
Support valueLink and onChange in combination.

colllin:


Andarist:
In the meantime `valueLink` got deprecated by React, and im planning to remove support for it from this lib entirely with the next major version. Therefore merging this for this short time is undesired. Gonna close it, thanks for the PR though.

PR #90, opened 2016-03-23
'none' is not a valid value for min-height

trun:
https://developer.mozilla.org/en-US/docs/Web/CSS/min-height

Here's an example of a site that has a global textarea style that sets min-height and ignores the invalid inlined min-height style.

developer_tools


andreypopp:
Thanks

PR #97, opened 2016-04-22
Remove requestAnimationFrame

zhm:
Using requestAnimationFrame makes the textarea not work correctly
in IE11 with React 15 in controlled mode. It causes the textarea to drop keystrokes
when typing quickly. The current demo site is using React 14 which doesn't seem to have the issue. Upgrading the demo page to React 15 causes the problem to happen.

Demo gif of the problem typing `hahahahahahaha` here:

ie11

I'm assuming the rAF code was added for a very specific reason or else it wouldn't be there, so I'm just opening this PR as a demo to show the problem with a fix.

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -63,7 +63,6 @@ export default class TextareaAutosize extends React.Component {
       minHeight: -Infinity,
       maxHeight: Infinity
     };
-    this._onNextFrameActionId = null;
     this._rootDOMNode = null;
     this._onChange = this._onChange.bind(this);
     this._resizeComponent = this._resizeComponent.bind(this);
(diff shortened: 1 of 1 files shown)


andreypopp:
Yeah, thanks, I'll do exactly that. Though I want to remove support for uncontrolled mode and valueLink as well so submitting these changes in a separate PR.

slorber:
+1 having an issue with cursor jumping at the end of input (with React 14). I'm not sure yet it's related to rAF but it might be as I had the problem in the past with rAF

fresk-nc:
+1

any news?

ahomu:
+1
hope to solve this problem...

slorber:
Hey guyz if you want your problem to be solved I'm using a fork available here:     "react-textarea-autosize-fork-slorber": "4.0.1",

Andarist:
@zhm could o check with the latest version? it could have fixed ur issue

Andarist:
From what ive tested the latest version of the package does no longer suffer from the problem. If you still experience the problem, please post back to me. However for now Im going to close this PR as the issue seems resolved.

PR #100, opened 2016-05-05
Remove rAF, also support for uncontrolled mode and valueLink

andreypopp:
Not sure why we needed rAF there.

Also I think it's better to remove support for unconrolled mode and valueLink. (I think this makes code much simpler and easier to reason about). Thoughts?

### DIFF
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -44,7 +44,6 @@ export default function calculateNodeHeight(
     hiddenTextarea = document.createElement('textarea');
     document.body.appendChild(hiddenTextarea);
   }
-  console.log('calculateNodeHeight', value);
 
   // Copy all CSS properties that have an impact on the height of the content in
   // the textbox
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -37,9 +37,10 @@ let hiddenTextarea;
 
 export default function calculateNodeHeight(
     uiTextNode,
-    value = null,
+    value,
     useCache = false,
     minRows = null, maxRows = null) {
+  console.log('calculateNodeHeight', JSON.stringify(value));
   if (!hiddenTextarea) {
     hiddenTextarea = document.createElement('textarea');
     document.body.appendChild(hiddenTextarea);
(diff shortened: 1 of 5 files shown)


src/calculateNodeHeight.js
```diff
@@ -35,9 +35,12 @@ const SIZING_STYLE = [
 let computedStyleCache = {};
 let hiddenTextarea;
 
-export default function calculateNodeHeight(uiTextNode,
+export default function calculateNodeHeight(
+    uiTextNode,
+    value,
     useCache = false,
     minRows = null, maxRows = null) {
+  console.log('calculateNodeHeight', JSON.stringify(value));
```
insin: Leftover debug logging?

andreypopp: Thanks

andreypopp:
Hm... actually we might need to add rAF back to batch `resize` event?

julen:
I'm using autosize as an uncontrolled component because I'm altering the value being input and I need to hack around the cursor position (see https://github.com/facebook/react/issues/955#issuecomment-160831548 for reference). Would this PR mean I cannot use the component in such a context anymore? That'd be a bummer 😕

Keats:
Any update on that?

slorber:
@andreypopp with release 4.0.2 and React 14, I had the Textarea to render twice, leading to the cursor going at the end of the input. This is because `_resizeComponent` was called twice (during both `_onChange` and `componentWillReceiveProps`). rAF may be a problem too in `componentWillReceiveProps`




Do you plan to release a fix for that soon?

slorber:
As a suggestion I think Om wraps all his inputs/textarea inside a custom component and avoid render them if value has not changed. This way it's not a big deal if you call setState too much on the parent or use rAF, leading to async rerenders of the input

slorber:
@andreypopp for the record, I just published the fork based on this branch: `react-textarea-autosize-fork-slorber@4.0.1`

This branch fixed my cursor problem because you avoid calling setState when only the input state changes:

```javascript
  componentWillReceiveProps(nextProps) {
    if (nextProps.value !== this.props.value) { // before it always trigger the resize
      this._resizeComponent(nextProps.value || nextProps.placeholder || '');
    }
  }
```

Please ping me when this is released so that I can stop using that fork and switch back to the original project :)

Keats:
@andreypopp need any help for that?

slorber:
@Keats help has been asked here: https://github.com/andreypopp/react-textarea-autosize/issues/115

PR #104, opened 2016-06-23
fix: removal of babel config from package.json

kgalli:
The babel setup used by this package does not work with different setups. This PR tries to fix this by using the `.babelrc` file for local development and ignoring it for the npm package.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+{
+  "presets": [
+    "prometheusresearch"
+  ]
+}
--- a/.npmignore
+++ b/.npmignore
@@ -2,3 +2,4 @@ src
 example
 Makefile
 npm-debug.log
+.babelrc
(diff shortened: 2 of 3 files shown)


andreypopp:
Thanks! Releases in 4.0.3.

SohKai:
@kgalli Oh man this was a serious headache for me today with `npm link` and babel-loader. Is it documented somewhere that a `package.json` config has different behaviour than a `.babelrc` config?

kgalli:
@SohKai what do you mean by different behaviour. In this case we do not need the preset config part in the npm package because it is already "compiled". But maybe I totally misunderstood your question?

SohKai:
I assumed that a babel config placed in either the `package.json` or a `.babelrc` would have the exact same effect, but this is clearly wrong given this pull request. I'm guessing that if the `package.json` has a config, that it gets used by babel plugins for module bundlers (ie. babel-loader)?

andreypopp:
Usually you want Babel to be configured to exclude all sources within node_modules (also speeds up build times). This is why I usually do Babel config in package.json and it's not a problem in that case.

PR #105, opened 2016-06-24
fix #91 retrieve textarea dom node

### LINKED ISSUE #91: Getting HTML element ref
Is it possible to get the element ref off of a textarea?

When doing something like this:

``` js
<TextArea ref='content' />
```

I'd expect to be able to do something like this:

``` js
this.refs.content.refs.content
```

which would get me the ref of the HTML element.

Unfortunately `this.refs.content.refs` returns an empty object.

In order to get the ref of the HTML element I have to do something like this:

``` js
this.refs.content._rootDOMNode
```

Let me know if I'm doing something wrong. Otherwise I'd like to figure out a way to get `refs` to work properly or make `_rootDOMNode` public.

vinnymac:
This exposes the `_rootDOMNode` to anyone who needs it, as explained [here](https://facebook.github.io/react/docs/more-about-refs.html).

 Example

``` javascript
import React from 'react';
import Textarea from 'react-textarea-autosize';

class Example extends React.Component {
  render() {
    <div onClick = {this._onClick}>
      <Textarea ref = 'area'/>
    </div>
  }

  _onClick(e) {
    let node = this.refs.area.getTextareaDOMNode();
    // do something special with the node here.
    // you can also use a function for the ref instead.
  }
}
```

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -205,6 +205,15 @@ export default class TextareaAutosize extends React.Component {
     this._rootDOMNode.blur();
   }
 
+  /**
+   * Convenience method to retrieve the textarea node.
+   * <Textarea ref = "area" />
+   * let node = this.refs.area.getTextareaDOMNode();
+   */
+  getTextareaDOMNode() {
+    return this._rootDOMNode;
+  }
+
 }
 
 function onNextFrame(cb) {


Andarist:
String `refs` are deprecated, and the preferred way to achieve such functionality is by a callback ref. Recently `inputRef` prop was added, so the consumers of the library can use it.

Gonna close this in favor of the `inputRef` prop.

PR #108, opened 2016-07-08
Set initial height to be any given props.style.height

dashed:
I'm using react-textarea-autosize for server-side rendering and the default style.height value is 0.

This patch enables user to pass their own props.style.height values for initial height.

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -59,7 +59,7 @@ export default class TextareaAutosize extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      height: null,
+      height: (props.style && props.style.height) || 0,
       minHeight: -Infinity,
       maxHeight: Infinity
     };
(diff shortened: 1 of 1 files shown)


jdrucey:
Any update on if this will be merged? This would help me solve the FOUC effect between the server rendering and the client rendering

stevelacy:
Just checking - would this be merged in the upcoming future?

Andarist:
@dashed wouldnt it be more appropriate if we put this in initial state in the constructor?

dashed:
@Andarist yeah, that can work too. 

I committed the appropriate changes; feel free to review. 👍

Andarist:
@dashed have u used ur fork in ur app? (even previous version of it) asking as confirmation that it didnt break anything would be nice before merging, although obviously it should be ok

dashed:
@Andarist I've used the initial PR-version previously; but haven't really tested the new version just committed recently.

I won't be able to test the new version until a week or so.

Andarist:
change is rather cosmetic, so Im gonna merge it in, thanks for the PR

Issue #112, opened 2016-07-19
Unknown props warning with react 15.2.1


afhole:
The non-standard attributes on the `textarea` generate warnings in React 15.2.1

```
Warning: Unknown props `minRows`, `onHeightChange`, `useCacheForDOMMeasurements` on <textarea> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
in textarea (created by TextareaAutosize)
```

Advice in React docs is to use custom attributes:

> You are using a non-standard DOM attribute on a native DOM node, perhaps to represent custom data. If you are trying to attach custom data to a standard DOM element, consider using a custom data attribute as described on MDN.

dashed:
For posterity, this warning persists in react v15.3.1.

Andarist:
Issue should be fixed for quite some time, if not - let me know and Im gonna reopen

PR #113, opened 2016-07-21
Fix 'react-unknown-prop'

psdcoder:
Fix to #112. Add blacklist of props which shouldn't pass to textarea.

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -6,6 +6,7 @@ import React from 'react';
 import calculateNodeHeight from './calculateNodeHeight';
 
 const emptyFunction = function() {};
+const blacklistProps = ['onHeightChange', 'useCacheForDOMMeasurements', 'minRows', 'maxRows'];
 
 export default class TextareaAutosize extends React.Component {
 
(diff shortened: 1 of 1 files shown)


richcsmith:
@andreypopp do you think we can get this merged?

psdcoder:
Hi, what status of PR?)

src/TextareaAutosize.js
```diff
@@ -92,9 +93,18 @@ export default class TextareaAutosize extends React.Component {
     if (maxHeight < this.state.height) {
       props.style.overflow = 'hidden';
     }
+
+    const filteredProps = Object.keys(props).reduce((acc, prop) => {
+      if (blacklistProps.indexOf(prop) === -1) {
```
hozefaj: Facebook recommends a somewhat cleaner solution https://facebook.github.io/react/warnings/unknown-prop.html

psdcoder: I think filtering better than deleting all unnecessary props from cloned object, it's more readable.
And there is only one right solution, isn't it?)

jdalton:
Let's go with the recommended approach.

PR #124, opened 2016-11-19
chore(*): Install rollup and build umd

rsnara:
It looks like the build files assume that the consumer is using the CommonJS module format. I didn't see a UMD build, so this PR introduces one. I'm relying on Rollup to do the module bundling. 

Without the UMD build, using this module within ClojureScript wasn't really possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ TESTS = $(wildcard src/__tests__/*-test.js)
 BABEL_OPTS = \
 	--sourceMaps=inline
 
-build: $(LIB)
+build: $(LIB) build-min
 
 test::
 	@echo No tests...
@@ -15,7 +15,7 @@ lint::
 	@$(BIN)/eslint $(SRC)
 
 clean:
-	@rm -f $(LIB)
+	@rm -rf lib/
 
 sloc:
 	@$(BIN)/sloc ./src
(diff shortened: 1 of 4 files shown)


rsnara:
Any thoughts, @jdalton?

Andarist:
@rsnara Ive added `es` build which should be consumable by rollup, we can still add UMD build built by rollup, but I would ask you to rebase this PR against master branch, cause they have started to overlap a little bit

rollup.config.js
```diff
@@ -0,0 +1,21 @@
...
+export default {
+  entry: 'src/TextareaAutosize.js',
+  dest: 'lib/TextareaAutosize.min.js',
+  format: 'umd',
+  moduleName: 'TextareaAutosize',
+  plugins: [
+    babel({
+      babelrc: false,
+      presets: ['es2015-rollup', 'stage-1', 'react'],
+      exclude: 'node_modules/**',
```
Andarist: we can also add `plugins: ['external-helpers']` here

Makefile
```diff
@@ -35,6 +35,10 @@ publish: build
 	@git push --tags origin HEAD:master
 	@npm publish
 
+build-min: $(LIB)
+	@echo "building lib/TextareaAutosize.min.js"
+	@$(BIN)/rollup -c
```
Andarist: would prefer have input / output files given as input for a `rollup` here rather than configured in a config

rollup.config.js
```diff
@@ -0,0 +1,21 @@
+import babel from 'rollup-plugin-babel';
+
+export default {
+  entry: 'src/TextareaAutosize.js',
+  dest: 'lib/TextareaAutosize.min.js',
```
Andarist: destination file should be put to `dist` rather than to `lib` + we could built both minified and unminified versions of the file (based on `NODE_ENV` ofc)

Andarist:
I've tweaked ur PR a little bit and merged it in. Thanks a lot!

Issue #120, opened 2016-10-18
TypeError: Cannot read property 'getAttribute' of null when testing with jest


BenSwennen:
I have a UserNote component that uses a TextArea from react-textarea-autosize and it works when previewing it in browser. However when I test the UserNote component, I get following error:

``` javascript
    TypeError: Cannot read property 'getAttribute' of null

      at calculateNodeStyling (node_modules/react-textarea-autosize/lib/calculateNodeHeight.js:82:21)
      at calculateNodeHeight (node_modules/react-textarea-autosize/lib/calculateNodeHeight.js:31:31)
      at TextareaAutosize._resizeComponent (node_modules/react-textarea-autosize/lib/TextareaAutosize.js:146:55)
      at TextareaAutosize.componentDidMount (node_modules/react-textarea-autosize/lib/TextareaAutosize.js:89:12)
      at node_modules/react/lib/ReactCompositeComponent.js:264:25
      at measureLifeCyclePerf (node_modules/react/lib/ReactCompositeComponent.js:74:12)
      at node_modules/react/lib/ReactCompositeComponent.js:263:11
      at CallbackQueue.notifyAll (node_modules/react/lib/CallbackQueue.js:67:22)
      at ReactTestReconcileTransaction.close (node_modules/react/lib/ReactTestReconcileTransaction.js:37:26)
      at ReactTestReconcileTransaction.closeAll (node_modules/react/lib/Transaction.js:204:25)
      at ReactTestReconcileTransaction.perform (node_modules/react/lib/Transaction.js:151:16)
      at batchedMountComponentIntoNode (node_modules/react/lib/ReactTestMount.js:63:27)
      at ReactDefaultBatchingStrategyTransaction.perform (node_modules/react/lib/Transaction.js:138:20)
      at Object.batchedUpdates (node_modules/react/lib/ReactDefaultBatchingStrategy.js:63:19)
      at Object.batchedUpdates (node_modules/react/lib/ReactUpdates.js:98:20)
      at Object.render (node_modules/react/lib/ReactTestMount.js:120:18)
      at Object.<anonymous> (src/components/Notes/__tests__/UserNote.test.js:45:75)
      at process._tickCallback (internal/process/next_tick.js:103:7)
```

Is there something I am doing wrong?

rovansteen:
[...]

PR #127, opened 2016-12-01
Check root dom node before computation

khankuan:
Component is breaking is a test case where dom is not available.


khankuan:
Related #120

Andarist:
Merged in, thanks for the fix.

PR #128, opened 2016-12-05
Check if hiddenTextarea is attached to document

2rp:
Hi @jdalton,

Thanks a lot for working on this :)

I'm working on a project where I'm using Rails' Turbolinks (https://github.com/turbolinks/turbolinks), which replaces the body tag, when navigating between pages.

This causes the textarea (used for calculations) to be removed from the DOM. However, there's still a reference the the global variable (hiddenTextarea), so the calculations become invalid.

I've therefore added a check to see if the textarea is in the DOM before making the calculations.

Cheers,
Soeren


src/calculateNodeHeight.js
```diff
@@ -34,10 +34,12 @@ const SIZING_STYLE = [
 
 let computedStyleCache = {};
 let hiddenTextarea;
+const hiddenTextareaId = 'react-textarea-autosize-textarea-for-calculations';
 
 export default function calculateNodeHeight(uiTextNode,
     useCache = false,
     minRows = null, maxRows = null) {
+  hiddenTextarea = document.getElementById(hiddenTextareaId);
```
Andarist: better check would be against `hiddenTextarea.parentNode` than polluting the global environment with custom id, would u like to revisit the code and make appropriate changes so we can merge it in?

Andarist:
@2rp ive merged in ur branch (with my extra commit) but by accident ive destroyed ur fork repository (didnt know I can push there) and subsequent pushes (which would fix revert my changes) got rejected, because 'permission denied', which is really odd as it didnt stop me before pushing there in the first place

If you want, we can still fix it (I have correct version of ur fork on my local machine) - I think u need to unprotect ur master branch or something. Let me know how can I help

PR #130, opened 2017-01-31
Fix onChange behavior on ie11

adrianton3:
`onChange` doesn't trigger correctly on IE11 in react 15 for inputs/textareas. `onInput` seems to work fine - I tested in IE10, 11, Chrome, Firefox and Safari.
There is a fix in the working here https://github.com/facebook/react/pull/8575 - also a good source of info on related issues

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -95,7 +95,7 @@ export default class TextareaAutosize extends React.Component {
     return (
       <textarea
         {...props}
-        onChange={this._onChange}
+        onInput={this._onChange}
         ref={this._onRootDOMNode}
         />
     );


Andarist:
Do u know when this regression in React has happened? Wouldnt like to include this as its quite a big internal change. Are there any explored alternatives? Also - what behaviour exactly is it fixing? Could u provide a shareable example of pre and post PR behaviours?

Andarist:
Please also check against the newest version of the library, Im testing it right now with controlled input and everything (regarding `onChange`) seems fine on IE11 (and other browsers)

Andarist:
@adrianton3 ive tested the newest version on IE 11 and everything seems fine, if you still experience the issue - could u post a reproduced example so I can test and debug it? going to close this for now but feel free to post back to me

peterkeating:
Was experiencing this issue, was resolved by updating to the latest version (`4.3.2`).

Andarist:
@peterkeating thanks for reporting back!

PR #131, opened 2017-02-01
Do not call _resizeComponent before the component is even updated. It…

slorber:
…'s useless and may trigger additional renderings leading to cursor jumps when change is rendered async

I don't really understand why this was called directly. What's the point of computing the height just before applying the change (it will be computed later when component receive parent props so it's not really needed)

If your state is hold by a parent, and the call to setState triggers an async rendering, then this line will actually make the cursor jump at the end of the input

### DIFF
--- a/src/TextareaAutosize.js
+++ b/src/TextareaAutosize.js
@@ -137,7 +137,6 @@ export default class TextareaAutosize extends React.Component {
   }
 
   _onChange(e) {
-    this._resizeComponent();
     let {valueLink, onChange} = this.props;
     if (valueLink) {
       valueLink.requestChange(e.target.value);


slorber:
For those who want to use my fork until merged, use this:

>     "react-textarea-autosize-seb-fork": "4.0.5",

jdalton:
Hi @slorber!

I'm not really sure either. Can you produce a tiny repro with a before and after example?

slorber:
Hi

@jdalton I've already tried, but actually I'm really really sure to understand when React decices to render sync/async on setState, so was unable to produce a small repro case

slorber:
I suspect this line has been the cause of another problem making user unable to write in textarea on iOS/cordova webview. Using my fork seems to have made this problem disappear (can't really confirm)

eliseumds:
Any updates on this? This PR seems to work fine.

slorber:
I'm still using my work and it works fine for me

Andarist:
Having faith in you @slorber ! Merging in ;)

Andarist:
the change broke uncontrolled mode, going to release a quick fix by detecting controlled/uncontrolled in the constructor by inspecting `props` and based on this call conditionally `_resizeComponent` in `_onChange`

PR #135, opened 2017-03-25
Only update state when values have changed

cchanna:
I found during performance testing that a not-insignificant amount of time was being spent going through the react lifecycle for this component. This is a simple optimization that avoids ever involving the lifecycle if the calculated height hasn't changed.


Andarist:
Thanks for the PR, gonna release later when I handle other PRs too.

PR #136, opened 2017-03-29
Apply style attributes directly instead of setAttribute, to avoid CSP unsafe inline

latentflip:
Unfortunately using `el.setAttribute("style", "...")` is blocked by our content security policy as it is considered "unsafe-inline" css.

This PR switches to direct property updates with `el.style[property] = value`, similar to how react itself handles changes to `style` on react elements.


Andarist:
Thanks for the fix! Going to publish soon, just need to take a look at other PRs etc

PR #141, opened 2017-04-08
Migrated deprecated React.PropTypes

Andarist:


marlonbernardes:
Thanks @Andarist! This is the only dependency I have that still hasn't been updated. I hope this will get merged soon!

PR #145, opened 2017-04-18
added .select() method

finnfiddle:
Added method for selecting all contents of textarea.

[See here](https://www.w3schools.com/jsref/met_textarea_select.asp).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-textarea-autosize",
   "description": "textarea component for React which grows with content",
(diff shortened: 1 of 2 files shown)


Andarist:
There is available `inputRef` prop, which will allow you to get reference to the underlaying DOM node and call `select()` on it. I think its quite new available prop, so it could have not been available at the time when you have created ur PR. Please use it though if you need this functionality.

Gonna close this for now. If my reasoning is not right, please just raise your concerns and will debate further.

PR #158, opened 2017-06-23
fix for font-style - calculating height size

wokr:
Include font-style attribute for calculating height.


Andarist:
Please revert version change in the `package.json`. It should always be done just prior the publication to npm by the publisher (in this case - me).

Other than that it's good to be merged in, thanks for the PR!

wokr:
done

Andarist:
Published

PR #164, opened 2017-07-05
Add check for falsy hiddenTextarea

ablamunits:
in `calculateNodeHeight`, hiddenTextarea can be false if not running in browser context,
which can cause tests to fail when accessing undefined `style` property further in the function.

### DIFF
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -45,6 +45,13 @@ export default function calculateNodeHeight(
   minRows = null,
   maxRows = null,
 ) {
+
+  // hiddenTextarea can be false if not running in browser context,
+  // which can cause tests to fail when accessing undefined style property.
+  if (!hiddenTextarea) {
+    return null;
+  }
+
   if (hiddenTextarea.parentNode === null) {
     document.body.appendChild(hiddenTextarea);
   }


Andarist:
@ablamunits thanks for the PR, could u specify though when this is a problem? `calculateNodeHeight` shouldnt run in a non-DOM context and its guarded against that [here](https://github.com/andreypopp/react-textarea-autosize/blob/9beaa4afc48f8c8541bc32ab8d990e731723763c/src/index.js#L134)

ablamunits:
@Andarist I am using mocha along with [jsdom-global](https://github.com/rstacruz/jsdom-global) for testing. Perhaps the use of `jsdom-global` plays a role here, but I am not too sure.
It looks like in your code you are relying on an `undefined` reference to the element to decide whether the node height calculation should take place. However, I don't believe that `calculateNodeHeight`, as a function, should be tightly coupled with that use. Moreover, as it seems, some testing environments are handling the reference well. 

Please let me know if there is something I might be missing here.

Andarist:
Would be great if you provide a simple repo with reproduced issue, so I could check how it behaves there exactly.

ablamunits:
@Andarist 
Thanks, I will try to set up a quick repo with an example soon.

ablamunits:
Hi @Andarist 
I have composed a minimal example showing the test error in a separate repo [here](https://github.com/ablamunits/react-textarea-autosize-example).

In the example, I am trying to create a simple wrapper component around `Textarea`, called `MyCustomTextarea`. In my tests for the custom component, I am simply interested in validating that the `value` is rendered correctly and the `onChange` callback is called.

To see the error, please clone, install deps and test with `npm run test`.
The error does not reproduce when running tests with Karma: `npm run test-with-karma`

Thanks!

Andarist:
Ok, so I see what's going on here. The problem lays in the switch of the context you are making here. `react-textarea-autosize` is required while its still in a non-DOM environment, but you polyfill it **later** with `jsdom-global`. 

I dont think that adding just next defensive check is the solution here. Im open however for a proposal how we could restructure the code better. Ofc please raise ur concerns

ablamunits:
@Andarist 
Thank you for taking a look. Would it be a reasonable suggestion to perform the checkes of whether or not the code is running in a browser context on demand? Instead of upon require?

Andarist:
Personally I dont like such solution as the code is working just fine and the proposed change is purely for satisfying the testing edge case.

Maybe we could expose a function though which you could use in your tests to reassure that the `hiddenTextarea` is there before you create an instance of Textarea. Maybe this would help others with some other edge cases too, although I cant think of any right now.

ablamunits:
To be honest, I think that by doing that you will be exposing the users to your implementation details.. I don't think something like the existence of `hiddenTextArea` is the user's concern.
I love that this component is described as a _drop-in replacement_, and I agree that this is an edge case, but I think it is still a bug.

With that being said, sometimes the reasonable solution is also the simplest one. I would love to contribute and give a shot at any solution here that would not require exposing functions for the sake of testing.

Andarist:
I think this is already fixed by [this line](https://github.com/andreypopp/react-textarea-autosize/blob/6a149bf6ba107ca86b11698399ba7a5779c6b85b/src/index.js#L132). `calculateNodeHeight` shouldnt be called in non-browser environments. If the issue is still not fixed please open a new one.

PR #168, opened 2017-07-19
add focus function to focus the inner textarea DOM element

yaelhe:
Currently there's no way to focus the textarea without accessing `_rootDOMNode`. It used to have focus() function on version 4.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -164,4 +164,8 @@ export default class TextareaAutosize extends React.Component {
 
     callback();
   };
+
+  focus = () => {
+    this._rootDOMNode.focus();
+  };
 }


Andarist:
The method was removed in v5 as redundant, the same functionality can be achieved by accessing the ref.
```js
componentDidUpdate() {
  // some logic
  this.inputRef.focus()
}

getRef = ref => (this.inputRef = ref)

render() {
  return <TextareaAutosize innerRef={ this.getRef }>
}
```
Therefore there is no need for additional proxying methods.

anaumov:
Just want to mention it should be `inputRef` instead of `innerRef`.

Andarist:
@anaumov thanks for noticing, was writing this in a rush :)

PR #169, opened 2017-07-19
Append textarea if body doesn't contain it

EricSimons:
This library currently treats the textarea as a singleton element. This assumption is fine as long as your entire page is a React SPA.

However, if React components are being used in a non-SPA fashion (i.e. if React isn't managing the entire page's DOM structure, routing, etc), then it opens up a bug where the textarea element is available in-memory but isn't actually a child element of the body anymore.

This causes the height calculations to get borked, as the "height" of a non-displayed textarea reults in negative integers. The component then styles the actual input based on negative values which wrecks all functionality.

To repro, initialize this component, wipe out the entire body tag, then try try to initialize it again.

This one line fix simply checks whether or not the textarea is a child node of the body instead of checking for the parent node, which ensures that the textarea is always a "visible" element and therefore will return non-negative numbers.


Andarist:
In what scenario exactly `hiddenTextarea` gets moved in the DOM? What kind of script is causing this for you?

EricSimons:
Turbolinks w/ React on Rails, it blows out entire body tag on every page navigation                                                                                                                                                                        -EricSent from my BlackBerry KEYone                                                                                                                                                                            From: notifications@github.comSent: July 19, 2017 3:32 PMTo: react-textarea-autosize@noreply.github.comReply-to: reply@reply.github.comCc: eric@esft.com; author@noreply.github.comSubject: Re: [andreypopp/react-textarea-autosize] Append textarea if body doesn't contain it (#169)  In what scenario exactly hiddenTextarea gets moved in the DOM? What kind of script is causing this for you?

Andarist:
Thats why we reattach this to body when no parent is present. Should work fine unless some script moves the hiddenTextarea and make it children of some other node than body

EricSimons:
Not if body.innerHTML is set to a completely new string. Any previous DOM nodes will be orphaned with their parent still set as the body element, but they aren't actually in the current active DOM body, thus leading to negative values for height/etc (since it's not being rendered at all), hence this pull request.                                                                                                                                                                         -EricSent from my BlackBerry KEYone                                                                                                                                                                            From: notifications@github.comSent: July 19, 2017 4:01 PMTo: react-textarea-autosize@noreply.github.comReply-to: reply@reply.github.comCc: eric@esft.com; author@noreply.github.comSubject: Re: [andreypopp/react-textarea-autosize] Append textarea if body doesn't contain it (#169)  Thats why we reattach this to body when no parent is present. Should work fine unless some script moves the hiddenTextarea and make it children of some other node than body

Andarist:
Ive just tested it and the `parentNode` seems to be null after I set `document.body.innerHTML`. Maybe its some kind of inconsistency between browsers? On what browser did u test this? Could u share a reproduced issue on something like codesandbox.io?

Andarist:
Closing because the original author did not provide more information that would help us to move with this further. If the issue is still present please open a new one with more details about this.

PR #174, opened 2017-08-24
Prevent NS_ERROR_NOT_AVAILABLE exception in Firefox in iframe

TLadd:
Putting a try-catch around the style.getPropertyValue calls to prevent hitting a Firefox bug in an iframe where it can throw a NS_ERROR_NOT_AVAILABLE exception. I am hitting this error periodically in my logs.

Firefox Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1012413
Linked jquery bug: https://bugs.jquery.com/ticket/15098
jsfiddle from the above jquery bug demonstrating the behavior: http://jsfiddle.net/LLeLa/


Andarist:
I'll look into the issue soon, but even if it throws and we return null - whats an expected behaviour then?

TLadd:
index.js is currently handling a null return value by just not altering state https://github.com/andreypopp/react-textarea-autosize/blob/master/src/index.js#L147

I think that behavior would be fine here as well. To be clear, this isn't something that happens 100% of the time when in an iframe in Firefox. Most of the time, the resizing works as expected. There are just some edge cases, I believe, where in between the call to getComputedStyle and getPropertyValue, the values become unavailable and firefox throws this exception.

TLadd:
Ended up resolving this by just putting an error boundary around react-textarea-autosize and falling back to a normal textarea. Think that approach is better than a catch all like this.

Andarist:
Probably u could retry once with `react-textarea-autosize` before falling back to a normal textarea. Maybe we should add such thing to the lib itself, although it would mean dropping support for React15.

Issue #99, opened 2016-05-03
Typing is garbled in IE11


insin:
Typing "Typing a test message" (and messing it up the second time :smile_cat:) with react/react-dom@15.0.2 and react-textarea-autosize@4.0.1:

![react-textarea-autosize-ie11](https://cloud.githubusercontent.com/assets/226692/14979163/7736891c-1118-11e6-84e6-a5a5e0e58360.gif)

The event handler isn't doing anything asynchronous:

``` js
  handleTextChange(e) {
    this.setState({text: e.target.value})
  },
```

This only seems to happen when the component is controlled - working around it by using it without a `value` prop. as that suits my current use case.

andreypopp:
Hm.. tried on a VM with IE11 downloaed from modern.ie:

![ie11-react-textarea-autosize](https://cloud.githubusercontent.com/assets/30594/14983856/54185f20-1148-11e6-92b9-3a3c9a2abb38.gif)

insin:
Is the purpose of this line in `componentWillUpdate()` to resize the textarea when it's controlled and a new value is forced into it via a prop, rather than by using the textarea?

```js
this._onNextFrameActionId = onNextFrame(this._resizeComponent);
```

Commenting this out makes IE11 behave when typing normally - could it be a difference in how IE implements `requestAnimationFrame`?

andreypopp:
Can you try code in `controlled` branch? I removed rAF call there and simplified code to allow only controlled usage (also removed support for `valueLink` pattern).

insin:
Looks good to me in IE

![controlled-branch](https://cloud.githubusercontent.com/assets/226692/15041588/fd73f5c6-12b4-11e6-9079-6ea2dd3790a2.gif)

andreypopp:
Ok, I'll wait for feedback on #100 a little bit and then release it (or not :-)).

averyvery:
We also have users experiencing this issue, glad someone caught it!

averyvery:
Any updates on this? Is it ready to merge?

frontendphil:
I would also like to see that being merged / fixed :) I'll have to disable this feature now, because I cannot ship to customers otherwise.

xylude:
I had this problem in an application, not with just this module, but regular text inputs too. React made a change in v15 to how the change event is handled in IE11 and below - my solution was just to change the 'onChange' event to 'onInput'. After that everything worked just fine.

Andarist:
[...]

PR #170, opened 2017-07-28
fixes Invalid calling object error on IE

ppong:
This fixes issue where `clearNextFrameAction(this._onNextFrameActionId)` throws an `Invalid calling object` exception on IE if `this._onNextFrameActionId` is undefined

PR #175, opened 2017-08-30
fix RAF bindings on IE11

amireh:
refs GH-99

Maintaining a reference to "requestAnimationFrame" and friends
(including setTimeout) without binding them to the window object seems
to make IE angry with an exception similar to:

    TypeError: Invalid calling object

Binding the function references to window fixes the issue.


src/index.js
```diff
@@ -10,10 +10,13 @@ import uid from './uid';
...
-  window.requestAnimationFrame
-  ? [window.requestAnimationFrame, window.cancelAnimationFrame]
-  : [setTimeout, clearTimeout];
+const [onNextFrame, clearNextFrameAction] =
+  isBrowser && window.requestAnimationFrame
+    ? [
+      window.requestAnimationFrame.bind(window),
+      window.cancelAnimationFrame.bind(window),
+    ]
+    : [window.setTimeout.bind(window), window.clearTimeout.bind(window)];
```
Andarist: Library might be used outside of the browser context (SSR, tests), so binding with `window` is not the best idea. Better to just create a proxy functions like:
```js
(fn, ms) => setTimeout(fn, ms)
```

amireh: Done

amireh: edit: in fact i removed that part of the change altogether since it's not necessary (no browser context = no ie, no need)

amireh:
Any chance we can get this in soon? We'd rather not keep using a fork / github in npm.

Thanks!

Andarist:
@amireh could u provide steps for reproduction? also your configuration would be helpful + webpack etc versions

Ive just setup a simple create-react-app project, tested it on IE11 and I have no issue with this.

cainlevy:
Confirming that this patch solves the issue I'm experiencing with IE11.

Andarist:
Without having exact steps for reproduction this PR has **zero** chances of being merged in. 

I want to be sure that the change is needed and that the issue originates here in the library and not in a tooling like webpack.

Does the problem occurs in dev or prod mode? What version of webpack do u use? What type of source maps? Etc, etc.

cainlevy:
I was able to finally reproduce this with `create-react-app` by ejecting and changing `devtool` to `eval-source-map` in `config/webpack.config.dev.js`:

1. make something tiny with create-react-app and react-textarea-autosize
2. eject
3. change `config/webpack.config.dev.js` to use `devtool: 'eval-source-map'`
4. load app in IE11
5. attempt to type in the textarea

Andarist:
Seems to be the same case as [this](https://github.com/andreypopp/react-textarea-autosize/pull/170#issuecomment-330994808). 

I don't intend to support IE quirk in **development mode only**, especially if it only happens when using "eval" source maps.

juhoha:
@Andarist I'd ask you to reconsider. Amireh's changes conform to better practices anyway. Using a forked version here as well: a slow devtool for a 30mb bundle isn't feasible for one dependency. A problem this small spawning a long-lived fork just seems silly.

Andarist:
I don't think that it really conforms to better practices, those functions are this-less and can be passed around as reference safely - the fact that IE has it bugged doesn't change anything.

However after a consideration I can add a development branch of this code to support this broken 💩.

cainlevy:
A big part of the value proposition for browser libraries is abstracting across browser differences. These are the quirks that shouldn't waste anyone's time, whether in development or in production. It's an imperfect situation that we have collectively been trying to make the best of for over a decade.

If the only fix for this issue is within a development branch, then it creates a discoverability issue and will continue to waste developers' time.

At the very least, please consider updating the compatibility statement in the readme:

> This module supports IE9 and above.

Andarist:
As the issue appears to be only relevant for using `eval` in IE11 (and I thought we've also as community established that using `eval` should be avoided), I think it's a reasonable thing to make the development branch different than production one.

Feel free to raise your concerns further, I'm also just figuring this thing out and as english is not my mother tongue I think my statements might sound a little bit rude and harsh - if so, please do not take them as such, this is not my intention.

cainlevy:
Thanks for keeping the discussion open! Also, thanks for maintaining a useful open source project. I want to remain respectful of your time and your ability to decide what's best for the project.

Do you have any examples in mind of other projects with development branches? How do people discover the branch, configure an app to use the correct branch in the correct environment, and gain confidence that what they're deploying has been tested?

When you first suggested a development branch, I guessed it would behave like an official fork. I may have misunderstood.

Andarist:
Oh, by the development branch I mean something like this:
```js
if (process.env.NODE_ENV === 'development') {
  // ...
} else {
  // ...
}
```
It's a commonly used practice for creating DEV only code. `process.env.NODE_ENV` gets replaced during build process and dead code branches are later removed by minifiers.

cainlevy:
Thanks @Andarist, that sounds like it would solve the problem. I do hope no one is using `eval` in production. 😃

juhoha:
I thought you meant a git branch/fork as well. But yeah, flagging the functionality behind env would definitely solve the problem :+1:

Andarist:
Sorry @amireh that it took so long and that finally your PR didn't get merged. I've just wanted to make changes quickly, so I could release it right now.

PR #180, opened 2017-11-13
Use tab-size css property when calculating node size

tmilloff:
First PR please be gentle


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-textarea-autosize",
   "description": "textarea component for React which grows with content",
-  "version": "5.2.0",
+  "version": "5.2.1",
```
Andarist: updating `dist` files and the version is part of the publishing process, should not be done in PRs

Andarist:
Besides the minor comment, I guess this is good to go.

tmilloff:
Thank you! Apologies for running/commiting `dist`; if I submit a PR in the future I will refrain from doing so.

Andarist:
Could u revert last 2 commits and push the branch out?

Andarist:
Thanks for the PR!

PR #190, opened 2018-02-22
docs: describe a way to set focus

maciej-ka:
I needed to set autofocus and had to look in the source for a way.

Perhaps it would be nice to also add a `autoFocus` prop that would set focus in `componentDidMount`. If you like this idea, I would prepare a PR.

Warm greetings!

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,17 @@ https://andreypopp.github.io/react-textarea-autosize/
 
 ## FAQ
 
+### How to focus
+
+Get a ref to inner textarea:
+```js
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks! There were some ppl asking how to do this from time to time, so this is very much appreciated ❤️ 

As to `autoFocus` - it's HTML attribute, we pass through all received props (omitting just few that are meant for this library) to underlaying DOM `textarea`, so it actually works:
```js
<TextArea autoFocus />
```
https://codesandbox.io/s/r0nnoxx5pn

PR #191, opened 2018-02-23
docs: describe autoFocus

maciej-ka:
Thanks for the [tip](https://github.com/andreypopp/react-textarea-autosize/pull/190) with autoFocus.
Perhaps it could be in docs too...


Andarist:
Thanks :)

Issue #157, opened 2017-06-13
Assertion of `_rootDOMNode` doesn't catch `null`


kevinehosford:
I see the following issue testing components that rely on `react-textarea-autosize`:
```
    TypeError: Cannot read property 'style' of null

      at Window.getComputedStyle (node_modules/jsdom/lib/jsdom/browser/Window.js:371:19)
      at calculateNodeStyling (node_modules/react-textarea-autosize/lib/calculateNodeHeight.js:111:22)
      at calculateNodeHeight (node_modules/react-textarea-autosize/lib/calculateNodeHeight.js:43:21)
      at TextareaAutosize._this._resizeComponent (node_modules/react-textarea-autosize/lib/index.js:78:58)
      at TextareaAutosize.componentDidMount (node_modules/react-textarea-autosize/lib/index.js:140:10)
      at node_modules/react-test-renderer/lib/ReactCompositeComponent.js:265:25
      at measureLifeCyclePerf (node_modules/react-test-renderer/lib/ReactCompositeComponent.js:75:12)
      at node_modules/react-test-renderer/lib/ReactCompositeComponent.js:264:11
      at CallbackQueue.notifyAll (node_modules/react-test-renderer/lib/CallbackQueue.js:76:22)
      at ReactTestReconcileTransaction.close (node_modules/react-test-renderer/lib/ReactTestReconcileTransaction.js:36:26)
```

It seems like the following check should handle `null`: https://github.com/andreypopp/react-textarea-autosize/blob/master/src/index.js#L134

Andarist:
What version of the library do u use?

kevinhos:
v5.0.6

Andarist:
Could u share repository with reproduced issue? It would certainly speed up the fix for this.

Andarist:
@kevinhos any updates on this?

kevinhos:
I apologize. I've been struggling to find time to put together an example.

savovs:
Works fine for me on 4.0.5, when I tried upgrading I got this error:
`"TypeError: Cannot read property 'getHostNode' of null"`

When doing this:

```javascript
<Textarea
							className={error ? 'alert' : ''}
							value={value}
							onChange={onChange}
							type={type}
							name={field}
							placeholder={placeholder}
							ref={componentReference => this._input = componentReference}
						/>
```

Andarist:
@savovs could u reproduce on something like codesandbox.io ?

[...]

PR #195, opened 2018-03-08
Guard against `this._rootDOMNode === null`

SimenB:
Fixes (probably): #157

I have the same issue as reported in that issue without using shallow render - I use [`storyshots`](https://github.com/storybooks/storybook/tree/master/addons/storyshots) where I have no control over the rendering.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -139,7 +139,7 @@ export default class TextareaAutosize extends React.Component {
   };
 
   _resizeComponent = (callback = noop) => {
-    if (typeof this._rootDOMNode === 'undefined') {
+    if (!this._rootDOMNode) {
       callback();
       return;
     }


Andarist:
Thanks for the PR! Should [this](https://github.com/storybooks/storybook/tree/master/addons/storyshots#using-createnodemock-to-mock-refs) be used instead?

SimenB:
Yeah, I just found it :D Was about to post a link. And it works
.

```diff
diff --git i/test/storyshots.test.js w/test/storyshots.test.js
index a600114..ad3ac35 100644
--- i/test/storyshots.test.js
+++ w/test/storyshots.test.js
@@ -5,12 +5,23 @@
-import initStoryshots from '@storybook/addon-storyshots';
+import initStoryshots, {
+    snapshotWithOptions,
+} from '@storybook/addon-storyshots';
 
+import TextareaAutosize from 'react-textarea-autosize';
+
 initStoryshots({
     configPath: path.resolve(
         __dirname,
         '../packages/troika-docs/storybook-config'
     ),
+    test: snapshotWithOptions({
+        createNodeMock: element => {
+            if (element.type === TextareaAutosize) {
+                return document.createElement('textarea');
+            }
+        },
+    }),
 });
```
So yes it works, but is there any reason to just check for `undefined` instead of falsy?

Andarist:
I think by making `calculateNodeHeight` unreachable with [this](https://github.com/andreypopp/react-textarea-autosize/blob/a2032a8028f97976263557bec8001b2b102f1be0/src/index.js#L131) should fix ur issue. This is not yet released but I'm going to prepare a new release in following days.

PR #200, opened 2018-05-08
fixed height=undefined

Schubidu:
In use with https://github.com/emotion-js/emotion/ there is an `console.error` `Warning: 'NaN' is an invalid value for the 'height' css style property.`. To fix that, it shall always give a valid height instead of `undefined`.


Schubidu:
If anyone have the same Issue I just published a scoped npm-package https://www.npmjs.com/package/@schubidu/react-textarea-autosize

Andarist:
Could u share a repro case for this undesired behaviour?

Andarist:
Closing because original author did not provide enough information to handle this.

PR #205, opened 2018-06-29
allows calculateNodeHeight function to be accessed externally

lufego:
### Description
We use the library because we have a textarea which expands as user types longer content.
But there's an wrapper called `Collapsible` which literally toggles the text area to be open or closed
when the user clicks a button

````js
const MIN_ROW_COUNT = 1

<Collapsible isDefaultClosed={this.props.isDefaultClosed}>
  {({ isOpen, toggle }) => (
    <TextareaAutosize
      minRows={MIN_ROW_COUNT}
      maxRows={isOpen ? undefined : MIN_ROW_COUNT}
      useCacheForDOMMeasurements={true}
    />

    {(this.state.rowCount > MIN_ROW_COUNT || !isOpen) && (
      <Button
        onClick={toggle}
        label={this.props.intl.formatMessage(
          messages[isOpen ? 'collapse' : 'expand']
        )}
      />
    )}
  )}
</Collapsible>
````

#### The problem
There's an option called `isDefaultClosed`, which means the component will be closed
when mounted no matter the content. But I need to render the "Expand/Collapse" 
button only when the content is bigger than one row. At the moment I can't do it
since I pass the maxRows from outside, and therefore it will always change the "height" calculation
based on the passed rows, leaving me no option to rely on the innerComponent anymore.

#### Solution
Since the logic for the button is external and whatever I'm passing it is changing its internal values
I need to somehow have this externally. I could then check if `isDefaultClosed` is passed and
if the content breaks into more than one row (with external function) and so I can render the button

#### Why I need the solution proposed in this PR
Accessing this function will allow me to check what is the height of the content
beforehands and check if I should render the "Expand/Collapse" button or not
as well as avoid me to recreate the same logic in another function

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -171,3 +171,6 @@ export default class TextareaAutosize extends React.Component {
     callback();
   };
 }
+
+// Exports calculateNodeHeight function to be accessible externally
+export { default as calculateNodeHeight } from './calculateNodeHeight';

### CODE REVIEW

**Comment by lufego** [comment]
Thanks for your quick answer @Andarist 

But.. I can't really use it 🙁that's why:

The parent component takes care of displaying the "Expand/Collapse" button, as well as giving the `TextArea` its height.

So if I want to know WHEN to display the button, I need to know if the content of TextArea has more than 1 row. But, then I'm already giving it a measurement from the outside:

````js
<TextareaAutosize
  minRows={MIN_ROW_COUNT}
  maxRows={isOpen ? undefined : MIN_ROW_COUNT} <-- Here
  useCacheForDOMMeasurements={true}
/>
````

In that line, I basically say that if `isOpen` (coming from `Collapsible) is true, the textarea should expand automatically to the content (calculated internally). If false, set it to 1 row.

But then, when I set to 1 row, I force the internal calculation to calculate a height as if it had only one row. Therefore I can't use the onHeightChange function.

That's why I need to calculate the height of the content separately, and therefore have access to the `calculateNodeHeight` function


----

PS: here's a fluxogram to understand a bit more (or make it even more complicated):

image

**Comment by Andarist** [comment]
Well, u wont be able to call `calculateNodeHeight` before mount anyway - so it seems to me (also looking on ur fluxogram) that it should be possible to render ur button conditionally based on:
```js
onHeightChange = (height, { rowCount }) => {
  // calculate if u need to render a button based on rowCount + isOpen
}
```

U might need to add some `componentDidUpdate` for `isOpen` changes but it should be fairly easy. Or have I missed smth? Your example seems nicely described, so you should be able to create codesandbox explaining the scenario in which it's hard to render ur output appropriately.

lufego:
@Andarist here it goes: https://codesandbox.io/s/k32k98qqkv

Basically my problem is that I pass a `maxRow`, and if you go to this part in `calculateNodeHeight.js`file, you see:

````js
if (maxRows !== null) {
  maxHeight = singleRowHeight * maxRows;
  if (boxSizing === 'border-box') {
    maxHeight = maxHeight + paddingSize + borderSize;
  }
  height = Math.min(maxHeight, height);
}
````

so those values will be always return as if there's one row.

lufego:
Exposing the `scrollHeight` even if I pass `maxRows`, would also solve my problem! that would be great. 

> shouldnt it be just this? this.state.numOfRows > TextArea.MIN_ROW_COUNT

I also pass the `!isOpen` condition to render the button with "expand" text when its **collapsed**

Andarist:
You can get `scrollHeight` from the ref (u have to pass `inputRef` prop to the component) - that still sounds troublesome for your use case, so we can figure out better solution.

> I also pass the !isOpen condition to render the button with "expand" text when its collapsed

According to given fluxogram and its left branch this is not true - the button should be rendered only if there is more text than 1 row. So to meet those requirements I dont see how `|| !isOpen` is useful.

Andarist:
closed in favour of #206

PR #206, opened 2018-07-03
adds valueRowCount property

lufego:
as discussed in https://github.com/andreypopp/react-textarea-autosize/pull/205


src/calculateNodeHeight.js
```diff
@@ -96,6 +96,10 @@ export default function calculateNodeHeight(
   hiddenTextarea.value = 'x';
   const singleRowHeight = hiddenTextarea.scrollHeight - paddingSize;
 
+  // Stores the value's rows count rendered in `hiddenTextarea`,
+  // regardless if `maxRows` or `minRows` props are passed
+  const valueRowCount = Math.floor(height / singleRowHeight);
```
lufego: not sure about the naming here. Wether `fullRowCount` would make more sense

Andarist:
Would this satisfy your use case? The thing I'm worried about that you probably will have to depend on ref to read this info, because it's not tightly correlated to `height` so it might change without `onHeightChange` firing.

lufego:
Yes. This fixes our problem

> The thing I'm worried about that you probably will have to depend on ref to read this info, because it's not tightly correlated to height so it might change without onHeightChange firing.

Not sure if I understand. How would it change without me changing the value itself, and consequently triggering `onHeightChange `? Plus, I don't think it would be a problem since I only update the state of the parent component when `onHeightChange` is fired.

````js
handleHeightChange = (_, innerComponent) => {
    this.setState({
      textareaRowCount: innerComponent.rowCount,
      contentRowCount: innerComponent.valueRowCount,
    });
  };
````

Andarist:
It can change if u i.e. have reached `maxRows` or `minRows` but type more text (or delete lines in second case) - then you won't be notified about new `valueRowCount` because `height` won't change.

lufego:
but it should be fine at this point since I just need to know if the content is smaller or bigger than 1 row

Andarist:
I've decided also to `expose` this in `this.props.onChange` - https://github.com/andreypopp/react-textarea-autosize/commit/582b796b19941f819b9142541f0ebc984243e8f5

I need to figure out some bundlers inconsistencies before releasing this though, hopefully should take no more than few days.

lufego:
thanks @Andarist. We are looking forward for the release

Andarist:
I've released new major version, please check it out.

PR #222, opened 2018-10-03
Setup Travis with size-limit test

### LINKED ISSUE #221: Setup Travis with size-limit test

ganny26:
I have implemented travis to test size-limit, fixes #221


package.json
```diff
@@ -27,7 +27,8 @@
     "preversion": "npm run lint",
     "release:patch": "npm version patch && npm publish && git push --follow-tags",
     "release:minor": "npm version minor && npm publish && git push --follow-tags",
-    "release:major": "npm version major && npm publish && git push --follow-tags"
+    "release:major": "npm version major && npm publish && git push --follow-tags",
+    "test": "jest"
```
Andarist: actually i think there is no need for jest right now, `scripts/update-size.js` should require `.size-limit.js` and run `sizeLimit` with required config (this is no functionality change for now, just unifying the size-limit config)

`.size-limit.js` will be required automatically when running `size-limit` CLI tool and this is the one that should be used for now on travis, this is described here https://github.com/ai/size-limit#usage

scripts/update-size.test.js
```diff
@@ -0,0 +1,31 @@
...
+const pkg = require(path.join(root, 'package.json'));
+const file = path.join(root, pkg.browser[pkg.module]);
+const size = require('./update-size');
+
+describe('size-limit', () => {
+  it('check size to be object', () => {
+    return size
+      .sizeLimit(file, { ignore: ['react', 'prop-types'] })
+      .then(result => {
+        expect(result).toBeInstanceOf(Object);
```
Andarist: this test doesnt really test anything - we dont want to retest size-limit package and its API

package.json
ganny26: Ok @Andarist  let me update my commit!

scripts/update-size.test.js
```diff
@@ -0,0 +1,31 @@
...
+      });
+  });
+
+  it('check gzip size limit', () => {
+    return size
+      .sizeLimit(file, { ignore: ['react', 'prop-types'] })
+      .then(result => {
+        expect(result.gzip).toBe(1673);
```
Andarist: this also is not something that we want to test, this number is already written to the README and this test would fail with any change made to the source,  the only thing that we need to test is the threshold size specified in `.size-limit.js` and thats what is going to be tested with `size-limit` tool

Andarist:
Thanks for working on this! I left few comments for now - it's a great start though

ganny26:
> now

HI @Andarist I have updated my PR #222  as per comments !

scripts/update-size.js
```diff
@@ -24,3 +24,7 @@ sizeLimit(file, { ignore: ['react', 'prop-types'] }).then(bytes => {
   // eslint-disable-next-line no-console
   console.log(`\n📦  Measured weight: ${weight}.\n`);
 });
+
+module.exports = {
```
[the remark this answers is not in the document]
ganny26: Thank you @Andarist All changes updated 👍

Andarist:
As mentioned - `scripts/update-size.js` should require `.size-limit.js` and pass it as config to the sizeLimit call.

ganny26:
> As mentioned - `scripts/update-size.js` should require `.size-limit.js` and pass it as config to the sizeLimit call.

Hey @Andarist  updated file now it will require config from .size-limit.js

Did a some update since we are utilising values from config i thought it will be nice when size is greater than config it will log a message.Is this fine for you?

`if (weight > sizeConfig[0].limit) {
    // eslint-disable-next-line no-console
    console.log(`\n⚠️  Project is now larger than ${sizeConfig[0].limit}\n`);
  }
`

scripts/update-size.js
```diff
@@ -6,13 +6,16 @@ const sizeLimit = require('size-limit');
 const root = path.join(__dirname, '..');
 const readmePath = path.join(root, 'README.md');
 const readme = fs.readFileSync(readmePath, 'utf-8');
+const sizeConfig = require('../.size-limit');
 
-const pkg = require(path.join(root, 'package.json'));
-const file = path.join(root, pkg.browser[pkg.module]);
-
-sizeLimit(file, { ignore: ['react', 'prop-types'] }).then(bytes => {
+sizeLimit(sizeConfig[0].path, sizeConfig[0].ignore).then(bytes => {
```
Andarist: We only have a single item in sizeConfig array, so it would be better to unpack that first item right away when requiring, like this:
```js
const sizeConfig = require('../.size-limit')[0];
```
and use it across the file without needing to access first item all over the place,

from what I see the second argument to `sizeLimit` is the opts object, so you should actually pass whole sizeConfig there (already being an object, not an array, after u make the first requested change), because what needs to be passed there is
```js
{ ignore: ['prop-types'] } 
```
and you are passing now
```js
['prop-types']
```

That's also probably why the reported size in the README has increased so much.

scripts/update-size.js
```diff
@@ -6,13 +6,16 @@ const sizeLimit = require('size-limit');
...
   const weight = formatBytes(bytes.gzip);
 
+  if (weight > sizeConfig[0].limit) {
+    // eslint-disable-next-line no-console
+    console.log(`\n⚠️  Project is now larger than ${sizeConfig[0].limit}\n`);
```
Andarist: that's good! I would only make it a console.warn instead of console.log

scripts/update-size.js
ganny26: Updated sizeLimit config

scripts/update-size.js
ganny26: Updated to console.warn 👍

scripts/update-size.js
```diff
@@ -6,13 +6,16 @@ const sizeLimit = require('size-limit');
...
+const sizeConfig = require('../.size-limit')[0];
 
-const pkg = require(path.join(root, 'package.json'));
-const file = path.join(root, pkg.browser[pkg.module]);
-
-sizeLimit(file, { ignore: ['react', 'prop-types'] }).then(bytes => {
+sizeLimit(sizeConfig.path, sizeConfig).then(bytes => {
   const weight = formatBytes(bytes.gzip);
 
+  if (weight > sizeConfig.limit) {
```
Andarist: I havent noticed it before, but you are comparing strings here, u should compare raw bytes (so you should convert `sizeConfig.limit` to them, or find a package that is able to compare those units).

ganny26: @Andarist I have used now [Bytes] module(https://www.npmjs.com/package/bytes) to compare sizes, earlier **format-bytes.js** was used i removed that since we have option to convert values in bytes module itself.After removing reduces size to **1.67KB to 1.63KB**

scripts/update-size.js
```diff
@@ -1,17 +1,15 @@
...
 const path = require('path');
-const formatBytes = require('format-bytes');
 const sizeLimit = require('size-limit');
-
 const root = path.join(__dirname, '..');
 const readmePath = path.join(root, 'README.md');
 const readme = fs.readFileSync(readmePath, 'utf-8');
 const sizeConfig = require('../.size-limit')[0];
-
+const byte = require('bytes');
```
Andarist: while this work bidirectionally I find it confusing, would prefer using more explicit parse and format functions, exported by this package

scripts/update-size.js
```diff
@@ -1,17 +1,15 @@
...
 sizeLimit(sizeConfig.path, sizeConfig).then(bytes => {
-  const weight = formatBytes(bytes.gzip);
+  const weight = byte(bytes.gzip);
```
Andarist: this could be moved down (after the conditional if) and could be written as:
```js
const weight = byte.format(bytes.gzip);
```

scripts/update-size.js
```diff
@@ -1,17 +1,15 @@
...
 
-  if (weight > sizeConfig.limit) {
+  if (byte(weight) > byte(sizeConfig.limit)) {
```
Andarist: and that would become
```js
if (weight > byte.parse(sizeConfig.limit)) {
```

ganny26: since weight variable has 1.63KB to validate condition we need to parse again

`if (byte.parse(weight) > byte.parse(sizeConfig.limit)` 

OR

`if (byte(weight) > byte.parse(sizeConfig.limit))`

shall i go with first option

Andarist: sorry, ive written wrong version of the code, should be
```js
if (bytes.gzip > byte.parse(sizeConfig.limit)) {
  // eslint-disable-next-line no-console
  console.warn(`\n⚠️  Project is now larger than ${sizeConfig.limit}\n`);
}

const weight = byte.format(bytes.gzip);
```

ganny26: update PR with changes👍

Andarist:
Thank you very much! And sorry for nitpicking so much 😅

PR #223, opened 2018-10-11
Setup test environment, added initial tests

chuntley:
This PR adds a testing environment to this repo. It uses a combination of jest and enzyme for component unit testing. Getting high coverage with the unit tests should help a lot, and in the future we can expand on this to run something like headless Chrome for running acceptance tests and capturing screenshots.


package.json
```diff
@@ -28,7 +28,9 @@
     "release:patch": "npm version patch && npm publish && git push --follow-tags",
     "release:minor": "npm version minor && npm publish && git push --follow-tags",
     "release:major": "npm version major && npm publish && git push --follow-tags",
-    "size": "size-limit"
+    "size": "size-limit",
+    "test": "jest",
+    "test:dev": "jest --watch"
```
Andarist: I would prefer here
`"test:watch": "npm test -- --watch"`

.babelrc.js
```diff
@@ -16,5 +16,10 @@ module.exports = {
     ['@babel/proposal-object-rest-spread', { loose }],
     ['transform-react-remove-prop-types', { mode: 'unsafe-wrap' }],
     'macros'
-  ]
+  ],
+  env: {
```
Andarist: this section kinda duplicates what is already specified in main babel configuration, would be IMHO better to reuse what's already there - I have also seen babel's `env` property rather unintuitive (do u know how those env-specific plugins/presets work in relation to common presets/plugins?)

I would prefer adding `@babel/transform-commonjs-modules` as devDep and adding it to common plugins in such way
```js
const test = process.env.NODE_ENV === 'test'

module.exports = {
  // ... config
  plugins: [
    // ... common plugins
    test && '@babel/transform-commonjs-modules'
  ].filter(Boolean)
}
```

src/__tests__/index.test.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
```
Andarist: would u be willing to refactor this to use `react-testing-library` instead of enzyme?

Andarist:
Thanks ❤️ I've left some comments, but this is a great start!

.travis.yml
```diff
@@ -2,4 +2,5 @@ language: node_js
 node_js:
   - "8.11.3"
 script: 
-  - npm run size
\ No newline at end of file
+  - npm run size
```
Andarist: reversed order of those scripts makes more sense to me, would prefer running test first and then size

chuntley:
Going through your feedback right now, I'll post an update tonight!

package.json
chuntley: 👍

src/__tests__/index.test.js
chuntley: I have not used that one before, I looked over the documentation and it appears to be missing features such as prop/state modifying. Since this component relies on state modifications to run, I recommend a library like enzyme is used for testing it.

.babelrc.js
chuntley: I was trying to remove the config on `@babel/env` here so it would default to commonjs for the modules property.

src/__tests__/index.test.js
Andarist: that's kinda the point of using this library instead of enzyme, consumers of this component don't modify state directly, so I think the tests shouldnt either - every state change has to be triggered by some event or rerendering (due to props changing), such tests would give us more confidence than testing internal implementation

chuntley: Testing the internal implementation is going to give you much more confidence in making changes in the future, I've found that only testing user-facing functionality (and snapshots) makes it very easy to miss bugs. I can add tests to this that mount the component as well, to have a combination of unit and functional tests.

Andarist: I don't quite agree with you. It's a common opinion (with which I agree), that integration tests can give you way more confidence than unit tests (example read https://blog.kentcdodds.com/write-tests-not-too-many-mostly-integration-5e8c7fff591c).

It's ofc not always easy to write integration tests and I don't want you to write any here, but I feel like moving higher this spectrum ladder is better. And that by imitating the user and their interactions with the component would give us more confidence. Because of the nature of this component (high coupling to the DOM as we have to measure the width of the hidden textarea) I'm not entirely sure if this is going to be easy using jest+jsdom, but might be worth exploring.

This would still be in my eyes a unit test (unit being a component), I'd rather like to test its interfaces and outputs rather than internal implementation.

That's ofc my opinion and I respect yours. I'm also really thankful for this contribution ❤️ any tests are better than none. If you don't feel like working towards what I'm proposing then its OK, I'll be happy to merge the PR as is and creating a separate ticket about trying out migration to react-testing-library.

chuntley:
Switched from enzyme to react-testing-library. Unfortunately I'm not familiar with that library, but hopefully this gives you a good starting point on adding to it in the future!

Andarist:
Thanks for your contribution, it's great! I'll file an issue about writing more tests - we could also explore using puppeteer or cypress to write more tests in the future.

PR #228, opened 2018-10-30
Document props. Fixes #227.

### LINKED ISSUE #227: props not documented
The valid props that could be given ( like `minRows` ) is not documented on the README as well as on the NPM page. It can be seen only by goint to the examples. If okay, I would be glad to add the documentation myself.

danedavid:


Andarist:
Superb :tada:

PR #229, opened 2018-11-13
Remove the hiddenTextArea from the document.body after it has been measured

hassoncs:
We don't want the document.body to fill up with extra elements over time (memory leak), so this change removes the hidden textarea after it has been measured.

Test plan: Ran the tests, verified everything still works.

### DIFF
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -66,6 +66,7 @@ export default function calculateNodeHeight(
   const nodeStyling = calculateNodeStyling(uiTextNode, uid, useCache);
 
   if (nodeStyling === null) {
+    hiddenTextarea.parentNode.removeChild(hiddenTextarea);
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


Andarist:
This element is shared across all instances of `react-textarea-autosize` - I would hardly call it a memory leak risk. Im concerned that removing & appending it with each key stroke when writing fast might not be best for performance of the component.

hassoncs:
Ah, thank you, @Andarist, I didn't realize the element was shared. The problem we are facing is our test suite checks the number of elements in the `document.body` at the start and end of each test and fails if there's anything extra added in the `body`. The test using `react-textarea-autosize` leaves textarea element in the body, causing a failure. I totally understand why you'd want to leave the element in the `body` between key strokes, so we might need to change something on our side.

Thanks for commenting so quickly– I appreciate it!

Andarist:
The only thing that comes to my mind that we could do to help you a little bit is to expose a cleanup function, just for testing purposes, which you could invoke before checking the numbers of elements in `document.body`. Not sure if it helps you anyhow as you'd have to invoke it manually anyway.

hassoncs:
I was thinking something similar, some kind of cleanup function. I'll close this PR to get it off your table. I'll open a new one if I get a chance to explore the cleanup function option. Thanks again!

PR #242, opened 2019-04-12
Link BundlePhobia

oliviertassinari:
https://bundlephobia.com/result?p=react-textarea-autosize


Andarist:
The reason why I'm not linking to bundlephobia is that I don't believe their metric is accurate. It's not that the value given in README here is taken out of thin air or that it's outdated - it actually comes from a measuring script:
https://github.com/andreypopp/react-textarea-autosize/blob/2cd8abd82a76d3b97b17bc80c2350133b2828932/package.json#L34

PR #243, opened 2019-05-08
Proposal: Allow callers to avoid subpixel height values

andrewkshim:
Subpixel rendering behavior depends on the browser and often isn't perfect (sometimes creating misalignment in layouts). This is a proposal to allow users to pass an `avoidSubpixelHeight` prop that will avoid subpixel height values by `ceil`ing the computed height (I think it's better to `ceil`, as opposed to `round`, so we don't decrease the height and potentially cut off some of the content).

Totally understand if you'd rather not include this to avoid bloating the API.

Thanks for the work (and apologies if this has been discussed before).

### DIFF
--- a/src/calculateNodeHeight.js
+++ b/src/calculateNodeHeight.js
@@ -56,6 +56,7 @@ export default function calculateNodeHeight(
   useCache = false,
   minRows = null,
   maxRows = null,
+  avoidSubpixelHeight = false,
 ) {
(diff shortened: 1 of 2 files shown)


Andarist:
Hm, couldn't you work around this by tweaking your styles? It would be much better if you just wouldn't run into this at all, rather than applying a fixup on the computed value.

PR #246, opened 2019-07-03
fix: Check _ref is not null

chdsbd:
Sometimes _ref is null and will raise an exception when passed into calculateNodeHeight.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -127,6 +127,10 @@ export default class TextareaAutosize extends React.Component {
       callback();
       return;
     }
+    if (!this._ref) {
+      callback();
+      return;
+    }
 
     const nodeHeight = calculateNodeHeight(
       this._ref,


Andarist:
Hi! Thanks for the PR, but to merge it I would have to know when exactly the current code might fail.

chdsbd:
I had a sentry error occur because _ref was null. I don't know how it happened, but a null check would prevent it in the future. Maybe _onChange was called before ref (not sure if that's possible with react).

Andarist:
> a null check would prevent it in the future

It would, but it could also cover up a root cause and that's never good. If we don't know what test could be written to prevent a broken behavior in the future then we shouldn't "fix" this blindly.

sbdchd:
We've encountered this issue again; seems to occur very rarely

Here is the stack trace:

```
TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
  at calculateNodeStyling(../node_modules/react-textarea-autosize/dist/react-textarea-autosize.esm.browser.js:128:1)
  at calculateNodeHeight(../node_modules/react-textarea-autosize/dist/react-textarea-autosize.esm.browser.js:52:1)
  at _resizeComponent(../node_modules/react-textarea-autosize/dist/react-textarea-autosize.esm.browser.js:209:1)
  at apply(../node_modules/react-textarea-autosize/dist/react-textarea-autosize.esm.browser.js:289:1)
  at wrapped(../node_modules/raven-js/src/raven.js:364:1)
```
Screen Shot 2019-11-04 at 2 15 09 PM

Occurred on Chrome 77.0.3865

I converted part of `react-textarea-autosize` to typescript and there seems to be a type error:


<details open>

<summary>index.js</summary>

```typescript
import * as React from 'react'
import calculateNodeHeight, {
  purgeCache,
} from 'components/TextAreaAutosize/calculateNodeHeight'

const noop = () => {}

let uid = 0

interface ITextareaAutosizeProps {
  readonly inputRef: React.MutableRefObject<HTMLTextAreaElement | null>
  readonly style: React.CSSProperties
  readonly value: string
  readonly maxRows: number
  readonly minRows: number
  readonly useCacheForDOMMeasurements: boolean
  readonly onHeightChange: (a: number, b: TextareaAutosize) => void
  readonly onChange: (
    a: React.ChangeEvent<HTMLTextAreaElement>,
    b: TextareaAutosize
  ) => void
}

interface ITextareaAutosizeState {
  readonly height: number
  readonly minHeight: number
  readonly maxHeight: number
}

export default class TextareaAutosize extends React.Component<
  ITextareaAutosizeProps,
  ITextareaAutosizeState
> {
  static defaultProps = {
    inputRef: noop,
    onChange: noop,
    onHeightChange: noop,
    useCacheForDOMMeasurements: false,
  }

  _uid: number
  _controlled: boolean
  _resizeLock: boolean
  _resizeListener: () => void | null
  _ref: HTMLTextAreaElement | null
  rowCount?: number
  valueRowCount?: number

  constructor(props: ITextareaAutosizeProps) {
    super(props)
    this.state = {
      height: (props.style && parseInt(String(props.style.height), 10)) || 0,
      minHeight: -Infinity,
      maxHeight: Infinity,
    }

    this._uid = uid++
    this._controlled = props.value !== undefined
    this._resizeLock = false
    this._resizeListener = () => {}
    this._ref = null
  }

  render() {
    const {
      inputRef: _inputRef,
      maxRows: _maxRows,
      minRows: _minRows,
      onHeightChange: _onHeightChange,
      useCacheForDOMMeasurements: _useCacheForDOMMeasurements,
      ...props
    } = this.props

    props.style = { ...props.style, height: this.state.height }

    const maxHeight = Math.max(
      parseInt(String(props.style.maxHeight), 10) || Infinity,
      this.state.maxHeight
    )

    if (maxHeight < this.state.height) {
      props.style.overflow = 'hidden'
    }

    return <textarea {...props} onChange={this._onChange} ref={this._onRef} />
  }

  componentDidMount() {
    this._resizeComponent()
    // Working around Firefox bug which runs resize listeners even when other JS is running at the same moment
    // causing competing rerenders (due to setState in the listener) in React.
    // More can be found here - facebook/react#6324
    this._resizeListener = () => {
      if (this._resizeLock) {
        return
      }
      this._resizeLock = true
      this._resizeComponent(() => {
        this._resizeLock = false
      })
    }
    window.addEventListener('resize', this._resizeListener)
  }

  componentDidUpdate(
    prevProps: ITextareaAutosizeProps,
    prevState: ITextareaAutosizeState
  ) {
    if (prevProps !== this.props) {
      this._resizeComponent()
    }

    if (this.state.height !== prevState.height) {
      this.props.onHeightChange(this.state.height, this)
    }
  }

  componentWillUnmount() {
    window.removeEventListener('resize', this._resizeListener)
    purgeCache(this._uid)
  }

  _onRef = (node: HTMLTextAreaElement | null) => {
    this._ref = node
    const { inputRef } = this.props

    if (typeof inputRef === 'function') {
      inputRef(node)
      return
    }

    inputRef.current = node
  }

  _onChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
    if (!this._controlled) {
      this._resizeComponent()
    }
    this.props.onChange(event, this)
  }

  _resizeComponent = (callback = noop) => {
    if (!process.env.BROWSER && !this._ref) {
      callback()
      return
    }

    const nodeHeight = calculateNodeHeight(
      this._ref,
     // ^^^^^^^ Argument of type 'HTMLTextAreaElement | null' is not assignable to parameter of type 'HTMLTextAreaElement'.
    // Type 'null' is not assignable to type 'HTMLTextAreaElement'.ts(2345)
      this._uid,
      this.props.useCacheForDOMMeasurements,
      this.props.minRows,
      this.props.maxRows
    )

    if (nodeHeight === null) {
      callback()
      return
    }

    const { height, minHeight, maxHeight, rowCount, valueRowCount } = nodeHeight

    this.rowCount = rowCount
    this.valueRowCount = valueRowCount

    if (
      this.state.height !== height ||
      this.state.minHeight !== minHeight ||
      this.state.maxHeight !== maxHeight
    ) {
      this.setState({ height, minHeight, maxHeight }, callback)
      return
    }

    callback()
  }
}
```

<details open>
<summary>calculateNodeHeight.js</summary>

```typescript
// --snip--
export default function calculateNodeHeight(
  uiTextNode: HTMLTextAreaElement,
  uid: number,
  useCache = false,
  minRows: null | number = null,
  maxRows: null | number = null
) {
// --snip--
```

</details>


## Type Error

```typescript
const nodeHeight = calculateNodeHeight(
  this._ref,
  // ^^^^^^
  // Argument of type 'HTMLTextAreaElement | null' is not assignable to parameter of type 'HTMLTextAreaElement'.
  //   Type 'null' is not assignable to type 'HTMLTextAreaElement'.ts(2345)
  this._uid,
  this.props.useCacheForDOMMeasurements,
  this.props.minRows,
  this.props.maxRows
)

```

Andarist:
I have recently published v8 which simplifies the code a little bit and I hope it gets rid of this problem. If you encounter it again with v8 - please let me know.

PR #265, opened 2020-05-24
fix: export Props type

SimenB:
We wrap this component at work, and wanna provide our own props based on the abse ones. Exporting it makes it easier to access (and matches what the package on DefinitelyTyped did)

(I made this edit in GH's UI, so a bit hard to add the changeset requested by the bot)

### DIFF
--- a/.changeset/spotty-hornets-wait.md
+++ b/.changeset/spotty-hornets-wait.md
@@ -1,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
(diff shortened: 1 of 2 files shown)


Andarist:
I suspect (although I'm not entirely sure) that the second link (the one mentioning being a maintainer) would work for you as well.

It's not a problem though - I will add a changeset in a minute and gonna rename this type to `TextareaAutosizeProps`.

Andarist:
Thanks!

PR #266, opened 2020-05-24
Fix calculatation of height with box sizing 'border-box'

vlazh:
Remove additional adding paddingSize and border because already contains in height (via getContentHeight)


Andarist:
Thanks! I've screwed up this when refactoring. I've applied a slightly different fix for this based on v7 implementation because your proposed fix would calculate a too high height for border-box textareas with border or padding. Thanks a lot about raising this issue and preparing a PR with a potential fix  ❤️

Issue #269, opened 2020-06-01
`syle` prop type error


Haaxor1689:
If I just try to pass text area attributes (from `React.InputHTMLAttributes`) there is a type error that `height` property of `style` does not match.
```
type Props = React.InputHTMLAttributes<HTMLTextAreaElement>;
const Component: React.FC<Props> = props => <TextareaAutosize {...props} />;
```

Here is the full error:

```
Type '{ accept?: string | undefined; alt?: string | undefined; autoComplete?: string | undefined; autoFocus?: boolean | undefined; capture?: string | boolean | undefined; checked?: boolean | undefined; ... 279 more ...; css?: InterpolationWithTheme<...>; }' is not assignable to type 'Pick<TextareaAutosizeProps, "style" | "autoComplete" | "autoFocus" | "disabled" | "form" | "maxLength" | "minLength" | "name" | "placeholder" | "readOnly" | ... 262 more ... | "wrap">'.
  Types of property 'style' are incompatible.
    Type 'CSSProperties | undefined' is not assignable to type '(CSSProperties & Pick<CSSProperties, "height" | "width" | "translate" | "color" 
| "clipPath" | "filter" | "marker" | "mask" | "alignContent" | ... 753 more ... | "vectorEffect"> & { ...; }) | undefined'.
      Type 'CSSProperties' is not assignable to type 'CSSProperties & Pick<CSSProperties, "height" | "width" | "translate" | "color" | "clipPath" | "filter" | "marker" | "mask" | "alignContent" | ... 753 more ... | "vectorEffect"> & { ...; }'.
        Type 'CSSProperties' is not assignable to type '{ height?: number | undefined; }'.
          Types of property 'height' are incompatible.
            Type 'string | number | undefined' is not assignable to type 'number | undefined'.
              Type 'string' is not assignable to type 'number | undefined'.  TS2322
```

I don't feel like I should be getting an error there.

Andarist:
Why would you like to use `props.style.height`? This is being controlled by the `react-textarea-autosize` because it's the only way to implement "autosizing". I don't believe that you should ever pass this to this component, apart from providing an initial value when SSRing.

Haaxor1689:
I don't want to use it but when I pass textarea props to my wrapping components I get errors. Shouldn't then the better solution be to omit `heigth` prop as well and also export the props type so typescript users that make wrapping components can just use the exported type for passing down the props? I'll make those changes to my PR.

[...]

PR #270, opened 2020-06-02
Removed props.style.height type override

Haaxor1689:
Because of the error mentioned in #269 I removed the `height` type override. It seems that this prop is not used anywhere at all anyways and I don't see a reason why `height` should specifically be a `number` only and not a `string | number` as it is defined normally.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -7,9 +7,7 @@ import { noop } from './utils';
 type Style = Omit<
   NonNullable<JSX.IntrinsicElements['textarea']['style']>,
   'maxHeight' | 'minHeight'
-> & {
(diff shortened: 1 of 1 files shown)


richBeth:
It would be nice if someone could fix this. I am currently experiencing an issue with the overwritten height as well.

JanRuettinger:
I am experiencing the same issue.

ryancarville:
+1 here as well.  When I extend `TextareaHTMLAttributes<HTMLTextAreaElement>` with `TextareaAutosizeProps` I get the error.  Even when omitting the `style` prop from `textarea`

PR #277, opened 2020-06-11
Updated ref attribute

edi:
Since `inputRef` support has been deprecated with version `8`, I think it's time for the `README` to reflect it.

Regards


Andarist:
Thanks! I've missed this when updating the docs.

PR #283, opened 2020-06-24
style: match `single-quote` prettier rule

emdotem:
I'm not used to this prettier configuration, sorry. 

I'd suggest to set up [Danger.js](https://danger.systems/js/) in PR to run some static checks like prettier.


emdotem:
@Andarist this should fix the release ;-)

PR #284, opened 2020-06-30
feat: expose rowHeight from onHeightChange

### LINKED ISSUE #271: Expose row count information
Hi,

we're trying to migrate to v8 but we realized that the change in `onHeightChange` function arguments is causing us to miss on a key feature.

**Background**

In our UI components library we have a component called `MultilineTextInput`, which uses the resizable textarea component. One of the features of this component is that when there is value in more than one line, we show an expand/collapse button.

You can see a live example here: https://uikit.commercetools.com/?path=/story/components-inputs--multilinetextinput

To do that, we rely on the `valueRowCount` exposed by the component.

https://github.com/commercetools/ui-kit/blob/86a85e642d0d47089c3ddd7094974c58badbed56/src/components/inputs/multiline-text-input/multiline-text-input.js#L30-L40

In v8 though, the `onHeightChange` only exposes the `height` value.

Ideally, what we need is the internally calculated `rowHeight`

https://github.com/Andarist/react-textarea-autosize/blob/757355da9d9810b7855621b0093cf26e323d48cd/src/calculateNodeHeight.ts#L50

And to calculate the row count:

```js
const valueRowCount = Math.floor(height / rowHeight);
```

**Proposal**

I'm not sure what the best solution here would be. Maybe we can introduce a new callback prop that specifically exposes this value.

What do you think? Do you see any other solution?

I'm happy to work on a PR if we decide how to proceed.

Thanks!

emmenko:
Fixes #271

### DIFF
--- a/.changeset/fair-bobcats-matter.md
+++ b/.changeset/fair-bobcats-matter.md
@@ -2,4 +2,4 @@
 'react-textarea-autosize': minor
 ---
 
-Expose `{ rowHeight: number }` as second argument of the callback function `onHeightChange`. This is useful to construct custom behaviors according to the height values.
+Added `{ rowHeight: number }` as a second parameter to the `onHeightChange` callback. This is useful to construct custom behaviors according to the height values.
--- a/.changeset/fair-bobcats-matter.md
+++ b/.changeset/fair-bobcats-matter.md
@@ -1,0 +1,5 @@
+---
+'react-textarea-autosize': minor
+---
+
+Expose `{ rowHeight: number }` as second argument of the callback function `onHeightChange`. This is useful to construct custom behaviors according to the height values.
--- a/src/calculateNodeHeight.ts
+++ b/src/calculateNodeHeight.ts
@@ -1,6 +1,11 @@
 import { SizingData } from './getSizingData';
 import forceHiddenStyles from './forceHiddenStyles';
 
+export type CalculatedNodeHeights = {
+  height: number;
+  rowHeight: number;
+};
+
 let hiddenTextarea: HTMLTextAreaElement | null = null;
 
 const getHeight = (node: HTMLElement, sizingData: SizingData): number => {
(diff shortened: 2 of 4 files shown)


src/calculateNodeHeight.ts
```diff
@@ -61,5 +66,5 @@ export default function calculateNodeHeight(
   }
   height = Math.min(maxHeight, height);
 
-  return height;
+  return { height, rowHeight };
```
Andarist, line 69: could we make this a tuple? the final code would be slightly smaller 😉 TS 4.0 will come for labels for tuples so we'll be able to improve readability for this then

.changeset/fair-bobcats-matter.md
```diff
@@ -0,0 +1,5 @@
+---
```
Andarist, line 1: Could you also update the README?

src/calculateNodeHeight.ts
emmenko: Sure, I wasn't sure if you wanted to expose more values.

.changeset/fair-bobcats-matter.md
emmenko: The description of the `onHeightChange` in the table?

Andarist: Ye

.changeset/fair-bobcats-matter.md
```diff
@@ -0,0 +1,5 @@
...
+'react-textarea-autosize': minor
+---
+
+Expose `{ rowHeight: number }` as second argument of the callback function `onHeightChange`. This is useful to construct custom behaviors according to the height values.
```
Andarist, line 5:
```suggestion
Added `{ rowHeight: number }` as a second parameter to the `onHeightChange` callback. This is useful to construct custom behaviors according to the height values.
```

PR #287, opened 2020-07-15
Update peerDependencies in package.json

MichaelDeBoey:
https://twitter.com/SemanticRelease/status/1249795728253353986

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "test:watch": "npm test -- --watch"
   },
   "peerDependencies": {
-    "react": "^16.8.0"
+    "react": ">=16.8.0"
   },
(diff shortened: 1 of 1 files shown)


Andarist:
I strongly agree with the reasons outlined [here](https://github.com/jaydenseric/apollo-upload-client/pull/201#issuecomment-657916102). At this moment in time, we can't know if this package will be compatible with React 17 or any further major version down the road.

MichaelDeBoey:
@Andarist If it doesn't work, you can always publish a minor that's restricting it more

Andarist:
If it will work I will be able to do the same - publish a new minor allowing a broader range 😉

PR #292, opened 2020-07-28
Check if hiddenTextarea is in current body

### LINKED ISSUE #247: Resize Stops Working after Browser Back
First off, I wanna say thanks for this awesome module! I'm thoroughly happy with it. Now, the issue I'm experiencing might be specific to my project because I'm using `react-router` and `turbolinks`. After navigating through pages via browser back/forward, resizing no longer works.

I did some debugging and found out the cause. Sometimes, the browser creates a new `document.body`, but `hiddenTextarea` keeps pointing at old `body` (and hence causing memory leak. Not sure where this behavior is from, BFCache might be the culprit here).

The quick fix I found is to change the following statement in [`calculateNodeHeight.js`](https://github.com/andreypopp/react-textarea-autosize/blob/efc9bc8e843825038fa1c3e7b8e221b6c673b463/src/calculateNodeHeight.js#L60)
```JavaScript
  // Before change
  if (hiddenTextarea.parentNode === null) {
    document.body.appendChild(hiddenTextarea);
  }
  // After change
  if (!document.body.contains(hiddenTextarea)) { /.../ }
```
This way, you can always check for the latest `body` to see whether or not `hiddenTextarea` exists in the current `body` instance. There might be a better way, but this is a one-liner fix.

If you'd like, I can quickly make the change, add some tests, and submit a PR. Let me know what you think :)

mendelk:
In situations where the `document.body` was replaced, the check if `hiddenTextarea.parentNode === null` would return `false` even if in fact the _new_ `document.body` had no current `hiddenTextarea`, thus requiring a new one to be appended. The new logic checks if `!document.body.contains(hiddenTextarea)`, which should account for the _current_ `document.body`.

An alternative solution would be to check if `hiddenTextarea.isConnected`, which is probably more performant. Unfortunately, this property is not present in IE.

Yet another solution which _might_ be more performant at the expense of being more verbose, is to recursively check if `hiddenTextarea.parentNode === null` to confirm that it's in fact still in the _current_ DOM tree.

Fixes #247

### DIFF
--- a/src/calculateNodeHeight.ts
+++ b/src/calculateNodeHeight.ts
@@ -33,7 +33,7 @@ export default function calculateNodeHeight(
     forceHiddenStyles(hiddenTextarea);
   }
 
-  if (hiddenTextarea.parentNode === null) {
+  if (!document.body.contains(hiddenTextarea)) {
     document.body.appendChild(hiddenTextarea);
   }
 


derrickreimer:
I encountered this situation (using Turbolinks) and can confirm that this PR solves the problem in my case.

mendelk:
@Andarist Is there anything I can do to help get this merged? I get that you've got lots of other things vying for your attention!

PR #305, opened 2021-02-09
fix: replace useEffect with useLayoutEffect for React 17

### LINKED ISSUE #259: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
We get this error from time to time. Wasn't able to reproduce but our clients have bumped into this bug.

image

Happened recently on:
1. browserName: Chrome, browserVersion: 71.0.3578, locale: ru-RU, osName: Linux
2. browserName: Chrome, browserVersion: 79.0.3945, locale: sv-SE, osName: Windows 10.0

Any ideas why this is happening?

mxschmitt:
`useEffect` cleanup runs since React 17 asynchronously thats why it needs to be changed

We deployed this change to a web application with quite a lot of users and could verify that the `Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.` errors were gone.

Fixes #259

cc @Andarist

### DIFF
--- a/.changeset/clean-kiwis-boil.md
+++ b/.changeset/clean-kiwis-boil.md
@@ -2,4 +2,4 @@
 'react-textarea-autosize': patch
 ---
 
-replace useEffect with useLayoutEffect for React 17
+Moved internal `'resize'` listener to the layout effect since React 17 calls cleanups of regular effects asynchronously. This ensures that we don't ever try to access the already unmounted ref in our listener.
--- a/.changeset/clean-kiwis-boil.md
+++ b/.changeset/clean-kiwis-boil.md
@@ -1,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+replace useEffect with useLayoutEffect for React 17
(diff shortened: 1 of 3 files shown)


Andarist (commented):
Could you also move usage of the `useWindowResizeListener` to this branch of the code?
https://github.com/Andarist/react-textarea-autosize/blob/9ba98b9a3442f36a91eb523c5be257237640cfda/src/index.tsx#L93

It will avoid React warnings about using `useLayoutEffect` during SSR.

mxschmitt:
@Andarist like that?

.changeset/clean-kiwis-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+replace useEffect with useLayoutEffect for React 17
```
Andarist, line 5: Moved internal `'resize'` listener to the layout effect since React 17 calls cleanups of regular effects asynchronously. This ensures that we don't ever try to access the already unmounted ref in our listener.

Andarist:
Thanks!

PR #318, opened 2021-04-08
Bug fixes for FF(87.0) & 

N1KN:
Bug fixes with incorrect scrollHeight & incorrect work with the "padding" in the latest version of FF;

Example: *[click](https://db1c1.csb.app/)*
_Check it in latest FF_

### DIFF
--- a/src/calculateNodeHeight.ts
+++ b/src/calculateNodeHeight.ts
@@ -20,6 +20,21 @@ const getHeight = (node: HTMLElement, sizingData: SizingData): number => {
   return height - sizingData.paddingSize;
 };
 
+const getNumberOfLinesInText = (value: string, minLines: number, maxLines: number): number => {
+  var lineBreakRegExp = /\r?\n|\r/g;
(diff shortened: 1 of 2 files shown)


Andarist:
The fix is quite involved and my head is not in this game right now. This package is dependent upon by a lot of users so I'm a little bit cautious about merging this. Could you describe in detail what was happening and how your fix attempts to solve the problem? Also - from what I understand it's a fix for FF and this was already fixed in their nightly releases. Which versions of FF are affected by this problem? Is it only the current one or were previous versions also affected?

N1KN:
I check it in last FF stable.
Other browsers including **FF esr(78.9.0)** and **FF 86.0.1** are not affected by this bug.

And it's easier for me to show it in a 1 minute video:

https://user-images.githubusercontent.com/37019408/114013904-b841a100-9870-11eb-81b2-335d2c702f77.mp4

The solution does not change how the component works in other browsers.

---
But in general, I would leave this pull request pending for people facing this problem.

PR #323, opened 2021-05-20
Allow elements other than textarea

LandonSchropp:
This is a draft of what I was thinking for allowing elements besides `textarea` to be used. Given some of the complexity around `contentEditable`, I'm not sure this is even a good idea. I wanted to put it out for comments and thoughts, but it's not ready to be merged yet.


Andarist:
Have you checked if this works correctly in the browser?

LandonSchropp:
@Andarist I starting down that road, but I haven't fully vetted it yet. I got sidetracked, and I'll explore that soon.

I was hoping to get some feedback first. Does this seem like a reasonable change, or is this out of scope for this package?

LoganDark:
I don't think contentEditable even needs this - DOM elements resize to their contents by default, and you can just specify min-height and max-height manually. Right?

LandonSchropp:
My use cases changed and I no longer need this, so I'm going to close it.

@LoganDark I didn't test in detail, but I hope you're right! If so, that'd simplify things for me if I need to come back to this.

PR #326, opened 2021-05-25
Fix Height calculation is not accurate when using word-break

### LINKED ISSUE #325: Height calculation is not accurate when using word-break
Currently, if `word-break:all` is set, the hidden textarea isn't inheriting this style which causes the height calculation of the hidden textarea to be inaccurate.

Example: 

CSS:
```css
.some-class {
    font-family: NotoSansKR,sans-serif;
    font-size: 16px;
    line-height: 1.6;
    word-break: keep-all;
}
```

JSX:
```jsx
  ..
    render() {
      <div className={"some-class"}>
            <TextareaAutosize />
      </div>;
    }
  ..
```

Rendered HTML:
```html
<body>
    <div class="some-class">
      <textarea style="height: 1104px;"></textarea>
    </div>
    <!-- Hidden textarea is rendered here but does not have the  style word-break:all set -->
    <textarea tab-index="-1" aria-hidden="true" style="min-height: 0px !important; max-height: none !important; height: 0px !important; visibility: hidden !important; overflow: hidden !important; position: absolute !important; z-index: -1000 !important; top: 0px !important; right: 0px !important; letter-spacing: normal; line-height: 23.299760818481445px; font-family: NotoSansKR, sans-serif; font-weight: normal; font-size: 16.06879997253418px; font-style: normal; tab-size: 8; text-rendering: auto; text-transform: none; width: 229px; text-indent: 0px; padding: 0px; border-width: 0px; box-sizing: border-box;"></textarea>
</body>
```

I'll create a PR for it :)

karlingen:
This fixes #325 

* Include `'wordBreak'` in sizing style

### DIFF
--- a/.changeset/quick-pianos-exercise.md
+++ b/.changeset/quick-pianos-exercise.md
@@ -2,4 +2,4 @@
 'react-textarea-autosize': patch
 ---
 
-Include word-break property when calculating the height
+Account for `word-break` property when calculating the height.
--- a/.changeset/quick-pianos-exercise.md
+++ b/.changeset/quick-pianos-exercise.md
@@ -1,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+Include word-break property when calculating the height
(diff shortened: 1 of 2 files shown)


chrisleyva (approved):

karlingen:
@Andarist Can we please have this one prioritized? Thank you!

PR #353, opened 2022-10-18
fix: workaround for Firefox bug with scrollHeight

knownasilya:
See https://github.com/Andarist/react-textarea-autosize/issues/313

Can be removed once https://bugzilla.mozilla.org/show_bug.cgi?id=1795904 is fixed and browser target is no longer supported.

### DIFF
--- a/.changeset/loud-boats-judge.md
+++ b/.changeset/loud-boats-judge.md
@@ -2,4 +2,4 @@
 'react-textarea-autosize': patch
 ---
 
-Workaround for Firefox scrollHeight bug affecting textarea height on dynamic toggle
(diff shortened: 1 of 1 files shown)


noahtallen:
@Andarist, I think this might help with #366 as well. Would love to see a fix! :)

.changeset/loud-boats-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+Workaround for Firefox scrollHeight bug affecting textarea height on dynamic toggle
```
Andarist, line 5:
```suggestion
Workaround for Firefox `scrollHeight` bug affecting textarea's height on a dynamic toggle
```

Issue #346, opened 2022-06-07
Cursor not at the end of input when using autofocus with initial value


ArnaudBarre:
Repro: https://stackblitz.com/edit/vitejs-vite-b4qb54?file=src%2FApp.jsx&terminal=dev

Temporary solution: Uncomment FixedTextarea line

AielloChan:
```tsx
import React, {
  forwardRef,
  useRef,
  useImperativeHandle,
  useLayoutEffect,
} from "react";
import {
  TextareaAutosize,
  TextareaAutosizeProps,
} from "react-textarea-autosize";

const Textarea = forwardRef(
  (props: TextareaAutosizeProps, ref: React.Ref<HTMLTextAreaElement>) => {
    const { autoFocus, ...restProps } = props;
    const textareaRef = useRef<HTMLTextAreaElement>(null);

    useImperativeHandle(ref, () => textareaRef.current!, []);
    useLayoutEffect(() => {
      if (autoFocus && textareaRef.current) {
        const textarea = textareaRef.current;
        textarea.focus();
        // Focus to the end
        const length = textarea.value.length;
        textarea.setSelectionRange(length, length);
      }
    }, [autoFocus]);

    return <TextareaAutosize {...restProps} ref={textareaRef} />;
  }
);
```

PR #357, opened 2022-11-12
fix: Cursor not at the end of input when using autofocus with initial…

Slowhand0309:
# Description
Fixed to move the caret position to the end of the character when `autoFocus` is true at the timing of `onFocus`.

# Related Issues
#346 

# Screen Capture
|before|after|
|-----|-----|
|||

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -33,6 +33,7 @@ const TextareaAutosize: React.ForwardRefRenderFunction<
     maxRows,
     minRows,
     onChange = noop,
+    onFocus = noop,
     onHeightChange = noop,
     ...props
   },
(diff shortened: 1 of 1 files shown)


Andarist:
To the best of my knowledge - this is how a regular `textarea` works in browsers (at least in some of them). I don't want to change the native semantics beyond what this library was set to add on top of them (autosizing).

Slowhand0309:
@Andarist 

> To the best of my knowledge - this is how a regular textarea works in browsers (at least in some of them). I don't want to change the native semantics beyond what this library was set to add on top of them (autosizing).

Thanks for the reply.
I see. I understand. It would indeed change the native semantics...
Please close this Pull Request. 🙏

AielloChan:
How can we implement this "auto focus at the end" behaviour with custom code?

AielloChan:
@Andarist Is there any recommendations? Thanks

AielloChan:
```
import React, {
  forwardRef,
  useRef,
  useImperativeHandle,
  useLayoutEffect,
} from "react";
import {
  TextareaAutosize,
  TextareaAutosizeProps,
} from "react-textarea-autosize";

const Textarea = forwardRef(
  (props: TextareaAutosizeProps, ref: React.Ref<HTMLTextAreaElement>) => {
    const { autoFocus, ...restProps } = props;
    const textareaRef = useRef<HTMLTextAreaElement>(null);

    useImperativeHandle(ref, () => textareaRef.current!, []);
    useLayoutEffect(() => {
      if (autoFocus && textareaRef.current) {
        const textarea = textareaRef.current;
        textarea.focus();
        // 将光标定位到内容末尾
        const length = textarea.value.length;
        textarea.setSelectionRange(length, length);
      }
    }, [autoFocus]);

    return <TextareaAutosize {...restProps} ref={textareaRef} />;
  }
);
```

PR #362, opened 2023-02-06
Handle custom fonts loading

ArnaudRinquin:
We had an issue using this library with a custom font for the textarea. It turns out the initial resizing could be done before the font is loaded.

This PR brings both the fix for it and adds such a case in the examples.

The fix is rather simple: listen for fonts being loaded, the same way we listen for the window being resized, and resize when it is done.


src/hooks.ts
```diff
@@ -18,3 +18,20 @@ export const useWindowResizeListener = (listener: (event: UIEvent) => any) => {
     };
   }, []);
 };
+
+export const useFontsLoadedListener = (listener: () => any) => {
```
Andarist, line 22:
```suggestion
export const useFontsLoadedListener = (listener: (ev: Event) => void) => {
```

src/hooks.ts
```diff
@@ -18,3 +18,20 @@ export const useWindowResizeListener = (listener: (event: UIEvent) => any) => {
...
+  const latestListener = useLatest(listener);
+
+  React.useLayoutEffect(() => {
+    const handler: typeof listener = () => {
+      console.log('loaded!');
```
Andarist, line 27:
```suggestion
```

src/index.tsx
```diff
@@ -93,6 +97,7 @@ const TextareaAutosize: React.ForwardRefRenderFunction<
   if (typeof document !== 'undefined') {
     React.useLayoutEffect(resizeTextarea);
     useWindowResizeListener(resizeTextarea);
+    useFontsLoadedListener(resizeTextarea);
```
Andarist, line 100: To save some bytes I would prefer to have a single generic hook that would be able to handle both of those.

You can copy-paste my implementation from [here](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAVwM4FMAyBDGKnwGZQQhwBEyKAtADZY4wkDcAUKJLHAEooYDG+hxElG59GTJjACeYFHACqAOwDWCiAHcFAMQQK+wCArgBeOAAoAdJYxQA5kgBccDAskBtALoBKYwD4nL5glpWQBJBTwUKABRADcUBRgAFWCAHkTE6xsUGD8TdMzsuBQAD2wFABMkOABvJjg4AHoGuDVZYHDIuAIiOAAmOAg4qCoIDEqWgAtgHgmnYTgkBBkoOB4iEAMu6CK4hLgYApgq9v2J2QARAHkAWTgqYAAjOsbm1tIwKH1PmGAALxQSKdZHhgFBcAMFLIMFUYGcuqDwQY2lUMFQ1BhJFUNvMkDIeMAQTxnmNyrF4jA0MBcPFIqYpDJHO0IisAAoAGjglnMmQc-kknkcMQgwHKzHqTRash4zjgwBsqnmsNkqDWFQhUOOMAA5FVkAhUVRJHAAFbIeAYLoGh68JSnLBOKgjNRVa3zAAGuE+ChsbrgZ2Ezwlb3IcDdTM6LN6vsGnTdzkkvuhkrgwgNRvKBi18Gl8wwDwgCHgsKp+2CcBiqIQKEDzWc5WTGYUWZaziLECcPHx5XJactwCoTlLMicSCQcoUef7wCkqzOPCUHOEIBQIAenVh9rjLl9VKb5tH47zVFkMHbbshQzdxPKpN2FKpZVp9JQjI6rN6HK5PMc8YF5eFopMAAvs8AD8cAsnAABkCwwF6NjPI4F6RIEz5wGEzJkgkKTPPktjZGyuHJMOJRlOMnrtAhuShgYAAk1TpMEQG+qR8TjEoKCSBAeBwHhWQwGBEHWBgy7YGCKRRKUUC8DAaQZPhMCuG6dEMcRKDMe4HKKCo6haDoegGD4PiuAADO4iFwFhMCBHg+k-Js5CUtSkJQDh9R8YUrEVFUVnyfxhHuWpRSlGxVQYZEvmpB5ORMD4pjPAcCmONFAVDi+vFqal9zOZEjimCgd6OOF0R3nJhwcoxMg+N4Rh+EKIpMN4tT1Kq4I0NguBOY+KwmI5tC4KY2XdZ4YqcCIMDmH1XGFlEeARHwpimDVfjNfUqwGOCEx1seUCOM+3F3A+NI9WYBXLXc-X3jlUDmDwCBQMICT5TEI3iGt+yHNyN5WV1x10sEHJbRUO2ve9wgwPdhhLb4H0KeYS4xj9R0uf9MiA9tkSg3AQEch4r0gUwjnI7SajtBmagckIOB-ACHLPedq0FTWcAAHqgcBr1E9dpgZndy4JOYeAGEclMjGMlGNrTp0vTDjMxMzbMc8wQA)

ArnaudRinquin:
Thanks @Andarist for the review. I tried using the code you provided but it isn't valid in the project's TS version (`^3.9.3`).

Switching TS version in the TS playground link you provided reveals it:
Screenshot 2023-02-06 at 15 36 31

It works if you select TS 4.4+ and forward.

Should we bump needed TS version or change the code somehow?

Andarist:
I think that it's fine to bump the TS version used by the repository. It won't actually have any effect on public consumers as those particular types are only relevant internally.

ArnaudRinquin:
@Andarist I upgraded the dev dependencies, eslint, babel and prettier along TS to make the whole setup work.

Andarist:
It seems that Babel is not happy about this syntax (CI failed) - I think that it should handle this by now though. Could you recheck what is happening?

ArnaudRinquin:
Damn, I will.

ArnaudRinquin:
@Andarist I'm not to happy with the shenanigans I had to go through to get the tests pass: adding JSDom + mocking the listeners. It doesn't really hurt but it's not crazy elegant either.

ArnaudRinquin:
Hey @Andarist do you see any blockers for this PR? Happy to adjust

.changeset/tricky-jokes-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+Support custom fonts loading
```
Andarist, line 5:
```suggestion
Support automatic resizing when a custom fonts ends up loading
```

PR #388, opened 2023-08-30
fix: Type 'CSSProperties' is not assignable to type 'Style' error

preflower:
`react-textarea-autosize` Style type internal conflict
image
and `CSSProperties` default supports `height` property, so I think we can remove union type

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -15,9 +15,7 @@ type TextareaProps = React.TextareaHTMLAttributes<HTMLTextAreaElement>;
 type Style = Omit<
   NonNullable<TextareaProps['style']>,
   'maxHeight' | 'minHeight'
-> & {
-  height?: number;
-};
+>;
 
 export type TextareaHeightChangeMeta = {
   rowHeight: number;


Andarist:
Please share a test case that motivates this change.

preflower:
> 

This is a type error, I can't provide a test case, or do you want a minimal reproducible code repo?

Andarist:
Yes, a sample code that errors but shouldn't (or the other way around).

preflower:
just because I try to set height is string type;
Here is a mini repo: https://stackblitz.com/edit/vite-react-ts-dwvju5?file=src%2FApp.tsx

preflower:
> just because I try to set height is string type; Here is a mini repo: [stackblitz.com/edit/vite-react-ts-dwvju5?file=src%2FApp.tsx](https://stackblitz.com/edit/vite-react-ts-dwvju5?file=src%2FApp.tsx)

Although height should not be applied in `react-textarea-autosize` component...should use `minRow` and `maxRow` instead of it.

Andarist:
> Although height should not be applied in react-textarea-autosize component...should use minRow and maxRow instead of it.

☝️ this. `height` is only there to aid with SSR scenarios when we can't measure the DOM.

preflower:
> > Although height should not be applied in react-textarea-autosize component...should use minRow and maxRow instead of it.
> 
> ☝️ this. `height` is only there to aid with SSR scenarios when we can't measure the DOM.

Yeah, but current `height` only support `number` type, it also should be support `string` type

PR #397, opened 2024-10-10
Prevent display style overrides from sites

Oyveloper:
We had an issue where our product was embedded on customer sites. One customer set the display: none for all hidden textareas. This makes sure that the hidden textarea always has the `display: block` which is required for the resizing to work.

### DIFF
--- a/.changeset/two-turkeys-join.md
+++ b/.changeset/two-turkeys-join.md
@@ -1,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+Force display: block for the hidden textarea to prevent site overrides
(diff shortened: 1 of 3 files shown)


.changeset/two-turkeys-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'react-textarea-autosize': patch
+---
+
+Force display: block for the hidden textarea to prevent site overrides
```
Andarist, line 5:
```suggestion
Force `display: block` for the hidden textarea to prevent other styles from overriding it and thus breaking the resizing functionality
```

PR #414, opened 2025-03-13
fix failed to execute getComputedStyle on window

benjaminwaterlot:
Can confirm this patch solved the issue in production for us, so this is the PR replicating it.

CleanShot 2025-03-13 at 17 09 16@2x

Fixes https://github.com/Andarist/react-textarea-autosize/issues/411

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -101,10 +101,10 @@ const TextareaAutosize: React.ForwardRefRenderFunction<
     React.useLayoutEffect(resizeTextarea);
     useFormResetListener(libRef, () => {
       if (!isControlled) {
-        const node = libRef.current!;
-        const currentValue = node.value;
+        const currentValue = libRef.current!.value;
         requestAnimationFrame(() => {
-          if (libRef.current && currentValue !== node.value) {
+          const node = libRef.current;
+          if (node && currentValue !== node.value) {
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -104,7 +104,7 @@ const TextareaAutosize: React.ForwardRefRenderFunction<
         const node = libRef.current!;
         const currentValue = node.value;
```
Andarist, line 105:
```suggestion
        const currentValue = libRef.current!.value;
```

src/index.tsx
```diff
@@ -104,7 +104,7 @@ const TextareaAutosize: React.ForwardRefRenderFunction<
...
         requestAnimationFrame(() => {
-          if (currentValue !== node.value) {
+          if (libRef.current && currentValue !== node.value) {
```
Andarist, line 107:
```suggestion
          const node = libRef.current;
          if (node && currentValue !== node.value) {
```

Andarist:
@benjaminwaterlot could you double-check applied changes?

benjaminwaterlot:
@Andarist it looks correct yes (but to be honest I'm not 100% confident either, I did not spend much time looking into the code). Thanks for the reactivity in any case! 🙏

